PR #4, opened 2019-07-29
Remove osgi deps

bmvermeer:


bmvermeer:
- support for older eclipse version now possible

PR #19, opened 2020-09-16
fix: quote path to file during snyk test

pavel-snyk:
This PR fixes issue with whitespaces in Maven subprojects.


maxjeffos (approved):
Reviewed together with Pavel

PR #37, opened 2022-05-19
feat: ✨ add preference for Telemetry [ROAD-911]

bastiandoetsch:
### Description

The used environment variable is the same as the CLI, that's why it's a bit confusing:

CLI checks for `Disablement`, while we check for `Enablement` (opt-in vs. opt-out).

This means, we invert the value of our preference when we populate the environment and
use the binary representation (`0|1`) for boolean.

The setting is `disabled` by default.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated
- [x] README.md updated, if user-facing

### Screenshots / GIFs

image


bastiandoetsch:
depends on https://github.com/snyk/snyk-eclipse-plugin/pull/35

CalamarBicefalo (approved):

PR #56, opened 2022-07-08
feat: add binary management capabilities

CalamarBicefalo:
# Breaking change

This release introduces a namespace for snyk LSP extensions, and therefore breaks compatibility with old LS servers.

## PR
- Adds ability to configure the CLI path
- Adds ability to disable automatic install/updates of binaries (LS & CLI)
- Uses the LS provided CLI as opposed to fetching another one


plugin/src/main/java/io/snyk/eclipse/plugin/SnykStartup.java
```diff
@@ -53,16 +50,6 @@ public void earlyStartup() {
       @Override
       protected IStatus run(IProgressMonitor monitor) {
         try {
-          File cliFile = getCliFile();
-          if (!cliFile.exists()) {
```
bastiandoetsch, line 57: `disableRunAbortActions` should be called here, and the actions should be enabled on receiving/having a valid CLI Path.

plugin/src/main/java/io/snyk/eclipse/plugin/properties/PreferencesPage.java
```diff
@@ -72,13 +57,17 @@ protected void createFieldEditors() {
     addField(new StringFieldEditor(Preferences.ADDITIONAL_PARAMETERS, "Additional Parameters:", getFieldEditorParent()));
     addField(new StringFieldEditor(Preferences.ADDITIONAL_ENVIRONMENT, "Additional Environment:", getFieldEditorParent()));
 
+    addField(space());
+    addField(new BooleanFieldEditor(Preferences.MANAGE_BINARIES_AUTOMATICALLY, "Update and install Snyk binaries automatically", getFieldEditorParent()));
+    addField(new FileFieldEditor(Preferences.LS_BINARY_KEY, "Snyk LS Path:", getFieldEditorParent()));
```
bastiandoetsch, line 62: Please make this `Snyk Language Server Path`. Feedback from customers was that they don't really know what `LS` stands for.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -31,13 +32,19 @@ public SnykExtendedLanguageClient(ProgressManager pm, Preferences preferences) {
     this.preferences = preferences;
   }
 
-  @JsonNotification(value = "$/hasAuthenticated")
+  @JsonNotification(value = "$/snyk.hasAuthenticated")
```
bastiandoetsch, line 35: Could we add an additional JsonNotification here for the old message name?

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -31,13 +32,19 @@ public SnykExtendedLanguageClient(ProgressManager pm, Preferences preferences) {
...
   public void hasAuthenticated(HasAuthenticatedParam param) {
     preferences.store(Preferences.AUTH_TOKEN_KEY, param.getToken());
     showAuthenticatedMessage();
     enableSnykViewRunActions();
   }
 
+  @JsonNotification(value = "$/snyk.hasDownloadedCli")
+  public void hasDownloadedCli(HasDownloadedCliParam param) {
```
bastiandoetsch, line 43: Is this message also sent, if a CLI is already existent, so that `enableSnykViewRunActions()` is also called if no download is needed?

plugin/src/main/java/io/snyk/eclipse/plugin/utils/FileSystemUtil.java
```diff
@@ -19,18 +20,18 @@ public final class FileSystemUtil {
    * @return File
    */
   public static File getCliFile() {
-    return new File(getCliDirectory(), Platform.current().snykWrapperFileName);
+    return new File(new Preferences().getCliPath());
```
michelkaporin, line 23: This seem to fail unit tests, since preferences mock doesn't get passed here. Can we address this?

PR #59, opened 2022-07-14
refactor: authentication

CalamarBicefalo:
### Description

Unifies token management by including it in the LSP exchange settings.

These settings are sent during LSP initialise and can be modified by the backend.


plugin/src/main/java/io/snyk/eclipse/plugin/EnvironmentConstant.java
```diff
@@ -0,0 +1,8 @@
+package io.snyk.eclipse.plugin;
+
+public interface EnvironmentConstant {
```
bastiandoetsch, line 3: I would name it in plural. But minor.

plugin/src/main/java/io/snyk/eclipse/plugin/SnykStartup.java
```diff
@@ -123,7 +109,14 @@ private boolean downloadLS() {
...
+  boolean isDownloadAllowed(Preferences preferences) {
+    return isDownloadAllowed(preferences.getLsBinary(), preferences.isManagedBinaries());
+  }
+  boolean isDownloadAllowed(String customPath, boolean managedBinaries) {
+    if (!managedBinaries) {
+        logger.info("LS: Automatically managed binaries disabled, not downloading");
+        return false;
+      }
     boolean customPathSet = customPath != null && !customPath.isBlank();
     if (customPathSet) {
```
bastiandoetsch, line 121: This is wrong - this couples managed binaries with custom path - a custom path does not mean that download is not automatic, it just means the automatic download/update should use the custom path as destination path

bastiandoetsch: I think this also means that we need changes to LSDownloader, to take the custom path as destination path.

plugin/src/main/java/io/snyk/eclipse/plugin/properties/PreferencesPage.java
```diff
@@ -30,22 +32,6 @@ public void init(IWorkbench workbench) {
 
   @Override
   protected void createFieldEditors() {
-    File lsFile = new LsRuntimeEnvironment(PREFERENCES).getLSFile();
-    String msg = "Snyk Language Server not found at " + lsFile.getAbsolutePath();
```
bastiandoetsch, line 34: The default path is never displayed anywhere anymore, which is a regression. It should probably be displayed in the Snyk LS path field. Clients have already needed and used that information a lot for troubleshooting and deleting LS from that location to trigger a fresh download.

plugin/src/main/java/io/snyk/eclipse/plugin/properties/PreferencesPage.java
```diff
@@ -71,13 +56,17 @@ protected void createFieldEditors() {
     addField(new StringFieldEditor(Preferences.ADDITIONAL_PARAMETERS, "Additional Parameters:", getFieldEditorParent()));
     addField(new StringFieldEditor(Preferences.ADDITIONAL_ENVIRONMENT, "Additional Environment:", getFieldEditorParent()));
 
+    addField(space());
+    addField(new BooleanFieldEditor(Preferences.MANAGE_BINARIES_AUTOMATICALLY, "Update and install Snyk binaries automatically", getFieldEditorParent()));
+    addField(new FileFieldEditor(Preferences.LS_BINARY_KEY, "Snyk LS Path:", getFieldEditorParent()));
```
bastiandoetsch, line 61: This should be `Language Server Path` or `Path to Language Server`

bastiandoetsch: Could also be `Snyk Language Server` to use the same naming scheme as `Snyk CLI Path`. But LS is a short acronym that we shouldn't use here.

plugin/src/main/java/io/snyk/eclipse/plugin/utils/FileSystemUtil.java
```diff
@@ -19,18 +20,18 @@ public final class FileSystemUtil {
    * @return File
    */
   public static File getCliFile() {
-    return new File(getCliDirectory(), Platform.current().snykWrapperFileName);
+    return new File(new Preferences().getCliPath());
```
bastiandoetsch, line 23: If I remember correctly, this broke Eclipse tests, because of not allowing injecting a Preference mock.

plugin/src/main/java/io/snyk/languageserver/SnykStreamConnectionProvider.java
```diff
@@ -59,4 +60,17 @@ protected ProcessBuilder createProcessBuilder() {
     runtimeEnvironment.updateEnvironment(pb.environment());
     return pb;
   }
+
+  @Override
+  public Object getInitializationOptions(URI rootUri) {
```
bastiandoetsch, line 65: Can we have a test for this?

plugin/src/main/java/io/snyk/eclipse/plugin/EnvironmentConstant.java
CalamarBicefalo: done

plugin/src/main/java/io/snyk/eclipse/plugin/SnykStartup.java
CalamarBicefalo: Done, that was indeed a hangover from the past. 

I don't think we need to change LSDownloader as it uses the path as defined in preferences.

bastiandoetsch:
Snyk Docs needs an update in the config section to have it sync'ed to the Readme.

plugin/src/main/java/io/snyk/eclipse/plugin/properties/PreferencesPage.java
CalamarBicefalo: done

plugin/src/main/java/io/snyk/eclipse/plugin/utils/FileSystemUtil.java
CalamarBicefalo: I have rewritten the preferences altogether. Constructor is now private, uses a static factory instead

plugin/src/main/java/io/snyk/languageserver/SnykStreamConnectionProvider.java
CalamarBicefalo: Hey, so I think since this method is just delegating, the valuable test here would be an integration test at the LSP layer.

I am not too clear on how to do that, I have asked in the lsp4e GH, if you have some ideas, happy to listen.

bastiandoetsch: Hm... a unit test would be fine I guess (could also be on LsConfigurationUpdater) just checking that we fill the settings object correctly.

tests/src/test/java/io/snyk/eclipse/plugin/SnykStartupTest.java
```diff
@@ -1,19 +1,27 @@
...
-    SnykStartup cut = new SnykStartup();
-    cut.setLogger(mock(ILog.class));
-    assertTrue(cut.isDownloadAllowed(""));
-    assertTrue(cut.isDownloadAllowed(null));
-    assertFalse(cut.isDownloadAllowed("a"));
+    SnykStartup startup = new SnykStartup();
+    startup.setLogger(mock(ILog.class));
+    assertTrue(startup.isDownloadAllowed("", true));
+    assertTrue(startup.isDownloadAllowed(null, true));
+    assertFalse(startup.isDownloadAllowed("a", true));
```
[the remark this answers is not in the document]
bastiandoetsch: I may be blind, but it's still `assertFalse`, right? So, that would assert the wrong thing - if a CLI path is set to `a` and the plugin should manage updates & downloads, it should download to that location. If the test succeeds, something is wrong.

plugin/src/main/java/io/snyk/languageserver/SnykStreamConnectionProvider.java
bastiandoetsch: Saw that you added it to LsConfigurationUpdater... resolving, thus.

plugin/src/main/java/io/snyk/eclipse/plugin/SnykStartup.java
```diff
@@ -123,18 +112,8 @@ private boolean downloadLS() {
...
-      logger.info("LS: Custom LS path is set, not downloading. Path: " + customPath);
-      return false;
-    }
-    return true;
-  }
-
   LsDownloader getLsDownloader() throws URISyntaxException {
-    URI baseUri = URI.create(LsDownloadRequest.getBaseURL());
-    IProxyData[] proxyData = runtimeEnvironment.getProxyService().select(baseUri);
+    IProxyData[] proxyData = runtimeEnvironment.getProxyService().select(LsBinaries.getBaseUri());
```
bastiandoetsch, line 116: Yeah, you can use asset URI there.

plugin/src/main/java/io/snyk/eclipse/plugin/properties/PreferencesPage.java
```diff
@@ -1,83 +1,71 @@
...
-  private static final Preferences PREFERENCES = new Preferences();
-
   public PreferencesPage() {
     super(GRID);
   }
 
   @Override
   public void init(IWorkbench workbench) {
-    setPreferenceStore(PREFERENCES.getStore());
+    setPreferenceStore(io.snyk.eclipse.plugin.properties.preferences.Preferences.getInstance().getStore());
```
bastiandoetsch, line 24: why is it not imported but referenced fully qualified? any reason?

plugin/src/main/java/io/snyk/languageserver/download/LsBinaries.java
```diff
@@ -0,0 +1,16 @@
+package io.snyk.languageserver.download;
+
+import java.net.URI;
+
+public class LsBinaries {
+  private static final String LS_DOWNLOAD_BASE_URL = "https://static.snyk.io/snyk-ls";
+  public static final String REQUIRED_LSP_VERSION = "2";
+
+  public static URI getBaseUri() {
```
bastiandoetsch, line 9: Can, but need not be ditched.

bastiandoetsch (approved):
Looks good to me. Please make sure the `rogue` import is removed so the tests pass and that the Snyk docs are updated with the configuration option changes, so the description matches the preference names. Bonus points: Update screenshots in gitbook ;)

PR #66, opened 2022-07-27
feat: add support for `window/showDocument` request [ROAD-1012]

bastiandoetsch:
### Description

To provide server controlled navigation, the `window/showDocument` message is needed. This PR adds support for it by implementing it. Next step is a PR to LSP4e to add it to the base plugin.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs
Navigation via code lenses possible through this PR.
image

### DIFF
--- a/plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
+++ b/plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
@@ -71,6 +71,7 @@ public class SnykExtendedLanguageClient extends LanguageClientImpl {
     ServerMessageHandler.showMessage("Authentication with Snyk successful", messageParams);
   }
 
+  // TODO: remove once LSP4e supports `showDocument` in its next release (it's been merged to it already)
   @Override
   public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams params) {
     return CompletableFuture.supplyAsync(() -> {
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -64,4 +70,19 @@ private void showAuthenticatedMessage() {
     messageParams.setMessage("The authentication token has been stored in Snyk Preferences.");
     ServerMessageHandler.showMessage("Authentication with Snyk successful", messageParams);
   }
+
+  @Override
+  public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams params) {
```
michelkaporin, line 75: Shouldn't we simply update to the latest LSP4E to get this? Or is the release cycle long there?

michelkaporin (approved):

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
michelkaporin: Perhaps worth noting it here, to wipe this code out later.

bastiandoetsch: Yeah, we should note it - release cycle is way slower than ours, but I'll add a comment.

PR #98, opened 2022-10-13
feat: update authentication user flow [ROAD-1198]

j-luong:
### Description

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

https://user-images.githubusercontent.com/16223568/198269390-1df6a831-792f-4cfe-bb74-8ef7fe4b0a49.mov


plugin/src/main/java/io/snyk/eclipse/plugin/wizards/SnykWizard.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+  
+  @Override
+  public String getWindowTitle() {
+    return "Snyk Wizard";
+  }
+  
+  @Override
+  public void addPages() {
+    configureAPI= new SnykWizardConfigureAPI();
```
bastiandoetsch, line 33: I'd add `Page` to the classname

plugin/src/main/java/io/snyk/eclipse/plugin/wizards/SnykWizardAuthenticate.java
```diff
@@ -0,0 +1,76 @@
...
+import org.eclipse.swt.widgets.Listener;
+
+public class SnykWizardAuthenticate extends WizardPage implements Listener {
+  private Text endpoint;
+  private Button unknownCerts;
+
+  public SnykWizardAuthenticate() {
+    super("Snyk Wizard");
+    setTitle("Authenticate");
+    setDescription("Review the endpoint configuration, clicking 'Finish' will authenticate with Snyk; this will open a new browser window.");
```
bastiandoetsch, line 25: Is this text confirmed by UX?

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -39,7 +50,27 @@ public static SnykExtendedLanguageClient getInstance() {
...
-    ExecuteCommandParams params = new ExecuteCommandParams("snyk.workspace.scan", new ArrayList<>());
+    // run the wizard if no token
+    if (Preferences.getInstance().getAuthToken().isBlank()) {
+      SnykWizard wizard = new SnykWizard();
+
+      WizardDialog dialog = new WizardDialog(PlatformUI.getWorkbench().getDisplay().getActiveShell(), wizard);
+
+      dialog.setBlockOnOpen(true);
+      dialog.open();
+    } else {
```
bastiandoetsch, line 61: Is it really an else? I think scan should be run after the wizard has finished, as that's the user intention.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -85,7 +116,8 @@ private void showAuthenticatedMessage() {
     ServerMessageHandler.showMessage("Authentication with Snyk successful", messageParams);
```
bastiandoetsch, line 116: This may be redundant, as we already send a notification from LS. Please check :)

plugin/plugin.xml
```diff
@@ -51,6 +51,11 @@
       id="io.snyk.eclipse.plugin.commands.execute"
       name="%command.name">
     </command>
+    <command
```
bastiandoetsch, line 54: Didn't we say that the wizard should be folded into the Snyk scan command instead of a second menu option?

bastiandoetsch: Also, the text of the menu entry should change IIRC, `Configure Snyk` instead of `Snyk Scan` when token is set. This can be achieved with a condition here.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
j-luong: It can be done without the else, the problem is that we will attempt a scan after clicking `finish` in the Wizard, however `finish` will initiate the authentication user flow. This results in a scan being attempted before the authentication flow is complete.

The alternative is to prevent the scan with the `else` or to complete the auth flow before the wizard finishes (which we intentionally changed to auth on clicking finish).

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
j-luong: Quick check seems to indicate that it's required (we don't get this notification when it is commented out)

plugin/plugin.xml
j-luong: We spoke to Annalisa who mentioned that a static `Snyk Test` over a dynamic `Configure Snyk / Snyk Test` was a better user experience.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
bastiandoetsch: Hm... we could trigger a scan after authentication, reacting to `$/snyk.hasAuthenticated`. That would be the event-driven approach.

bastiandoetsch (commented):
We still need to address triggering a scan after authentication but that should be easy to do in `SnykExtendedLanguageServerClient`

michelkaporin (approved):

PR #108, opened 2022-12-07
fix: null exception in wizard [ROAD-1244]

asaf92:
### Description

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

### DIFF
--- a/plugin/src/main/java/io/snyk/eclipse/plugin/wizards/SnykWizardModel.java
+++ b/plugin/src/main/java/io/snyk/eclipse/plugin/wizards/SnykWizardModel.java
@@ -9,9 +9,9 @@ public class SnykWizardModel {
   private String initialAuthToken;
 
   public SnykWizardModel() {
-    this.initialEndpoint = Preferences.getInstance().getPref(Preferences.ENDPOINT_KEY);
-    this.initialUnknownCerts = Preferences.getInstance().getPref(Preferences.INSECURE_KEY);
-    this.initialAuthToken = Preferences.getInstance().getPref(Preferences.AUTH_TOKEN_KEY);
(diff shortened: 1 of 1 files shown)


michelkaporin (commented):
Nice catch! Should we utilise proper getters for other ctor initializers as well?

michelkaporin:
@asaf92 maybe worth to unify and centralise this logic, instead of introducing complexity with two different flows?

plugin/src/main/java/io/snyk/eclipse/plugin/wizards/SnykWizardModel.java
```diff
@@ -9,9 +9,9 @@ public class SnykWizardModel {
   private String initialAuthToken;
 
   public SnykWizardModel() {
-    this.initialEndpoint = Preferences.getInstance().getPref(Preferences.ENDPOINT_KEY);
-    this.initialUnknownCerts = Preferences.getInstance().getPref(Preferences.INSECURE_KEY);
-    this.initialAuthToken = Preferences.getInstance().getPref(Preferences.AUTH_TOKEN_KEY);
+    this.initialEndpoint = Preferences.getInstance().getPref(Preferences.ENDPOINT_KEY, "");
+    this.initialUnknownCerts = Preferences.getInstance().getPref(Preferences.INSECURE_KEY, "");
```
michelkaporin, line 13: Having this as empty string can potentially lead to type errors in runtime, since this is supposed to be bool.

asaf92: Fixed

bastiandoetsch (approved):

PR #130, opened 2023-04-04
feat: use refresh token instead of snyk token

j-luong:
### Description

_This PR should implement oauth support via LS_

### Checklist

- [ ] Tests added and all succeed
- [ ] Linted
- [ ] CHANGELOG.md updated
- [ ] README.md updated, if user-facing

PR #132, opened 2023-04-12
feat: support oauth authentication flow

j-luong:
### Description

_This PR should implement oauth support via LS_

This PR replaces #130 due to incorrect branch naming.

### Checklist

- [x] Linted


tests/src/test/java/io/snyk/eclipse/plugin/properties/preferences/PreferencesTest.java
```diff
@@ -33,6 +33,7 @@ void test_DefaultPreferences() {
     assertEquals("true", prefs.getPref(Preferences.MANAGE_BINARIES_AUTOMATICALLY));
     assertEquals("true", prefs.getPref(Preferences.MANAGE_BINARIES_AUTOMATICALLY));
     assertEquals("1", prefs.getPref(Preferences.LSP_VERSION));
+    assertEquals("token", prefs.getPref(Preferences.AUTHENTICATION_METHOD));
```
bastiandoetsch, line 36: This should use the constant that defines the value "token" instead of using the string value "token" directly.

bastiandoetsch (approved):

PR #137, opened 2023-05-09
feat: when using OAuth set SNYK_OAUTH_TOKEN when calling the CLI

PeterSchafer:
### Description
When the IDE uses OAuth authentication, for the invocation of the CLI it shall provide the credentials via the SNYK_OAUTH_TOKEN variable which contains only the access token. This is enough, since the IDE will maintain the OAUTH token, including refreshing it.

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated


plugin/src/main/java/io/snyk/eclipse/plugin/runner/ProcessRunner.java
```diff
@@ -122,13 +127,18 @@ private void setupProcessBuilderBase(ProcessBuilder pb) {
 
     String authMethod = Preferences.getInstance().getPref(Preferences.AUTHENTICATION_METHOD);
     String token = Preferences.getInstance().getAuthToken();
-    if (token != null && authMethod.equals(Preferences.AUTH_METHOD_OAUTH)) {      
-      pb.environment().put(EnvironmentConstants.ENV_INTERNAL_SNYK_OAUTH_ENABLED, "1");
-      pb.environment().put(EnvironmentConstants.ENV_INTERNAL_OAUTH_TOKEN_STORAGE, token);
-      pb.environment().remove(EnvironmentConstants.ENV_SNYK_TOKEN);
+    if (token != null && authMethod.equals(Preferences.AUTH_METHOD_OAUTH)) {   
+      try {
+	      ObjectMapper objectMapper = new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
bastiandoetsch, line 132: If I remember correctly, there's already code somewhere to parse the OAuth token - could we maybe extract that code to remove duplication?

bastiandoetsch (approved):

PR #147, opened 2023-07-19
Fix: custom path for CLI

PeterSchafer:
### Description
remove incorrect fix-up of custom paths

### Checklist

- [ ] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs
N/A

### DIFF
--- a/plugin/src/main/java/io/snyk/languageserver/LsRuntimeEnvironment.java
+++ b/plugin/src/main/java/io/snyk/languageserver/LsRuntimeEnvironment.java
@@ -123,16 +123,9 @@ public class LsRuntimeEnvironment {
     var pathOptional = Preferences.getInstance().getPath();
     if (pathOptional.isPresent()) {
       var path = pathOptional.get();
-      String splitBy;
-      if (path.contains(";")) {
-        splitBy = ";";
-      } else {
-        splitBy = ":";
-      }
-      String newPath = path.replace(splitBy, File.pathSeparator)
-        + File.pathSeparator
+      String newPath = path + File.pathSeparator
         + System.getenv("PATH");
-      env.put("PATH", newPath);
+		env.put("PATH", newPath);
     }
   }
 
--- a/tests/src/test/java/io/snyk/languageserver/LsRuntimeEnvironmentTest.java
+++ b/tests/src/test/java/io/snyk/languageserver/LsRuntimeEnvironmentTest.java
@@ -18,6 +18,7 @@ import org.osgi.framework.ServiceReference;
 import org.osgi.framework.Version;
 
 import java.util.HashMap;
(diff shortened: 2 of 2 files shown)


tests/src/test/java/io/snyk/languageserver/LsRuntimeEnvironmentTest.java
```diff
@@ -145,4 +145,14 @@ void testEnableTelemetryIsAddedToEnvironmentDisabled() throws StorageException {
     // This is a bit confusing - CLI takes DISABLE as env variable, but we ask for ENABLE, so it's reverted
     assertEquals("1", env.get(Preferences.ENABLE_TELEMETRY));
   }
+  @Test
+  void testAddPath() throws StorageException {
+	String expected = "C;/myPath/:";
```
bastiandoetsch, line 150: tha'ts a beautiful path.

tests/src/test/java/io/snyk/languageserver/LsRuntimeEnvironmentTest.java
```diff
@@ -145,4 +145,14 @@ void testEnableTelemetryIsAddedToEnvironmentDisabled() throws StorageException {
...
+    HashMap<String, String> env = new HashMap<>();
+    when(preferenceMock.getPref(Preferences.PATH_KEY)).thenReturn(expected);
+    
+    environment.addPath(env);
+
+    assert(env.get("PATH").startsWith(expected));
```
bastiandoetsch, line 156: Well, wouldn't the path just be transmitted to LS, and not added to the env?

plugin/src/main/java/io/snyk/languageserver/LsRuntimeEnvironment.java
```diff
@@ -123,16 +123,9 @@ public void addPath(Map<String, String> env) {
...
-        splitBy = ";";
-      } else {
-        splitBy = ":";
-      }
-      String newPath = path.replace(splitBy, File.pathSeparator)
-        + File.pathSeparator
+      String newPath = path + File.pathSeparator
         + System.getenv("PATH");
-      env.put("PATH", newPath);
+		env.put("PATH", newPath);
```
bastiandoetsch, line 128: I wonder if this should be done.

tests/src/test/java/io/snyk/languageserver/LsRuntimeEnvironmentTest.java
PeterSchafer: While addPath() still lives at LsRuntimeEnviornment, it is now only used for CLI invocations. This is why it still needs to be fixed.

bastiandoetsch (approved):

PR #149, opened 2023-09-18
feat: enables SCLE by using LS to fetch SAST settings [nebula-1427]

metju90:
This PR deletes the `apiClient` and uses snyk-language-server to fetch SAST settings


bastiandoetsch (approved):
LGTM

PR #231, opened 2024-12-02
fix: replaced imageDescriptors

acke:
### Description

image

### Checklist

- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs


plugin/src/main/java/io/snyk/eclipse/plugin/utils/SnykIcons.java
```diff
@@ -22,4 +22,10 @@ public class SnykIcons {
 
 	public static final ImageDescriptor ENABLED = Activator.getImageDescriptor("/icons/enabled.png");
 	public static final ImageDescriptor DISABLED = Activator.getImageDescriptor("/icons/transparent.png");
+
+	public static final ImageDescriptor PROJECT = Activator
+			.getImageDescriptor("platform:/plugin/org.eclipse.ui.ide/icons/full/obj16/prj_obj.png");
+	public static final ImageDescriptor FILE = Activator
+			.getImageDescriptor("platform:/plugin/org.eclipse.ui.genericeditor/icons/full/obj16/generic_editor.png");
```
bastiandoetsch, line 29: this fixes the issue, but what if we have a file icon that is not the generic editor one?

bastiandoetsch: E.g. a Java file, or a Python file when PyDev is installed.

acke: How about we keep the original logic, but replace the return null, with return SnykIcons.FILE?

PR #255, opened 2025-01-30
feat: fix lint

acke:
### Description

This PR solves a number of lint errors.

### Checklist

- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs


plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
```diff
@@ -208,7 +209,7 @@ public String getAutomaticAuthentication() {
 		}
 
 		public String[] getTrustedFolders() {
-			return trustedFolders;
+			return Arrays.copyOf(trustedFolders, trustedFolders.length);
```
andrewrobinsonhodges-snyk, line 212: The local trustedFolders is already a copy of the input (given the change at line 131) - why copy it again? Also, why the change to using arrays.copyOf instead of clone (or new String(trustedFolders))? It feels like having to specify the length makes this more fragile.

plugin/src/main/java/io/snyk/languageserver/SnykIssueCache.java
```diff
@@ -81,7 +82,7 @@ public Collection<Issue> getIssues(String path, String displayProduct) {
 	 */
 	public void addCodeIssues(String path, Collection<Issue> issues) {
 		if (!Paths.get(path).startsWith(basePath)) {
```
andrewrobinsonhodges-snyk, line 84: We could move this out to a private function checkPath(String path) or similar to avoid all the duplication.

andrewrobinsonhodges-snyk (approved):
Happy with this overall, though there are a few inconsistencies and further optimisations - see the comments.

The only thing that needs addressing before merge is your local path in the ruleset file (plugin/.pmd) - the rest are up to you.

plugin/.pmd
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<pmd>
+    <useProjectRuleSet>true</useProjectRuleSet>
+    <ruleSetFile>/Users/skippy/repos/snyk-eclipse-plugin/plugin/src/main/resources/pmd-ruleset.xml</ruleSetFile>
```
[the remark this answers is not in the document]
bastiandoetsch: could be helpful to have the same rules as the CI/CD. But of course the absolute path is a bummer :(

plugin/.ruleset
bastiandoetsch, line 1: We should be able to automatically generate it from our "master" XML file in the root. But that's something for a different PR.

plugin/src/main/java/io/snyk/eclipse/plugin/views/ScanWorkspaceFolderHandler.java
```diff
@@ -19,6 +19,7 @@
 public class ScanWorkspaceFolderHandler extends AbstractHandler {
 
 	@SuppressWarnings("restriction")
+	@Override
 	public Object execute(ExecutionEvent event) throws ExecutionException {
```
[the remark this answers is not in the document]
bastiandoetsch: This was there before. I guess it can be fixed another day ;)

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/filters/BaseFilter.java
```diff
@@ -5,14 +5,14 @@
 import io.snyk.eclipse.plugin.preferences.Preferences;
 import io.snyk.eclipse.plugin.views.snyktoolview.TreeFilterManager;
 
-@SuppressWarnings("rawtypes")
-public abstract class BaseFilter {
+@SuppressWarnings({ "rawtypes"})
+public class BaseFilter {
```
bastiandoetsch, line 9: Why not abstract anymore? We don't want to instantiate it, right?

plugin/src/main/java/io/snyk/eclipse/plugin/wizards/SnykWizard.java
```diff
@@ -44,25 +44,29 @@ public void addPages() {
...
+	@Override
 	public void init(IWorkbench workbench, IStructuredSelection selection) {
 		this.workbench = workbench;
 		this.selection = selection;
 	}
 
+	@Override
 	public boolean canFinish() {
-		if (this.getContainer().getCurrentPage() == authenticatePage) {
+		if (this.getContainer().getCurrentPage().equals(authenticatePage)) {
```
bastiandoetsch, line 55: Good one: you found Kotlin syntax bug ;)

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
```diff
@@ -127,7 +128,7 @@ public Settings(String activateSnykOpenSource, String activateSnykCodeSecurity,
 			this.integrationName = integrationName;
 			this.integrationVersion = integrationVersion;
 			this.automaticAuthentication = automaticAuthentication;
-			this.trustedFolders = trustedFolders;
+			this.trustedFolders = trustedFolders.clone();
```
bastiandoetsch, line 131: I doubt the clone method is the right choice here. It probably, IIRC, does not clone the array/list.

bastiandoetsch: Decided to look it up https://www.perplexity.ai/search/arrays-clone-in-java-_mPlckijTZGLk9lBkR4UvA

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
bastiandoetsch: Yep, clone() seems to be [preferred](https://www.perplexity.ai/search/arrays-clone-in-java-_mPlckijTZGLk9lBkR4UvA).

plugin/src/main/java/io/snyk/languageserver/download/HttpClientFactory.java
```diff
@@ -30,70 +30,73 @@
...
+			} catch (NoSuchAlgorithmException | KeyStoreException | KeyManagementException e) {
+				SnykLogger.logError(e);
+			}
+		}
+		return httpClientBuilder.build();
+	}
+
+	IProxyData getRelevantProxyData(IProxyData... proxyData) {
+		for (IProxyData data : proxyData) {
+			if (data.getHost() != null) {
```
bastiandoetsch, line 68: Why invert it? I find it easier to understand, if the positive logic is not wrapped in an `if` statement and indented.

plugin/src/main/java/io/snyk/languageserver/protocolextension/ProgressManager.java
```diff
@@ -12,15 +12,15 @@
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 
 public class ProgressManager {
-	SnykExtendedLanguageClient lc = SnykExtendedLanguageClient.getInstance();
+	SnykExtendedLanguageClient lc;
```
bastiandoetsch, line 15: Why is this correct? This could leave the ProgressManager with a null instance of lc, doesn't it?

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -314,7 +314,7 @@ public void hasAuthenticated(HasAuthenticatedParam param) {
 		}
 
 		String newToken = param.getToken();
-		boolean differentToken = newToken != oldToken;
+		boolean differentToken = !newToken.equals(oldToken);
```
bastiandoetsch, line 317: Ouch. Nice find.

plugin/src/main/java/io/snyk/languageserver/protocolextension/messageObjects/PublishDiagnostics316Param.java
```diff
@@ -9,47 +9,52 @@ public PublishDiagnostics316Param() {
...
+	}
 
-    public Diagnostic316[] getDiagnostics() {
-        return diagnostics;
-    }
+	public Diagnostic316[] getDiagnostics() {
+	    if (diagnostics == null) {
+	        return new Diagnostic316[0]; // Return an empty array if diagnostics is null
+	    }
+	    return Arrays.copyOf(diagnostics, diagnostics.length);
```
bastiandoetsch, line 26: same as above - why not just use clone()?

plugin/src/main/java/io/snyk/languageserver/protocolextension/messageObjects/PublishDiagnostics316Param.java
```diff
@@ -9,47 +9,52 @@ public PublishDiagnostics316Param() {
...
+	public void setUri(String uri) {
+		this.uri = uri;
+	}
 
-    public Diagnostic316[] getDiagnostics() {
-        return diagnostics;
-    }
+	public Diagnostic316[] getDiagnostics() {
+	    if (diagnostics == null) {
+	        return new Diagnostic316[0]; // Return an empty array if diagnostics is null
```
bastiandoetsch, line 24: Why this? Isn't this a semantical change?

plugin/src/main/java/io/snyk/languageserver/protocolextension/messageObjects/PublishDiagnostics316Param.java
```diff
@@ -9,47 +9,52 @@ public PublishDiagnostics316Param() {
...
+	}
 
-    public void setDiagnostics(Diagnostic316[] diagnostics) {
-        this.diagnostics = diagnostics;
-    }
-    
-    @Override
+
+	public void setDiagnostics(Diagnostic316... diagnostics) {
+		this.diagnostics = diagnostics.clone();
```
bastiandoetsch, line 31: if empty, will be a nullpointerexception.

plugin/src/main/java/io/snyk/languageserver/protocolextension/messageObjects/SnykTrustedFoldersParams.java
```diff
@@ -1,13 +1,15 @@
 package io.snyk.languageserver.protocolextension.messageObjects;
 
+import java.util.Arrays;
+
 public class SnykTrustedFoldersParams {
   private String[] trustedFolders;
 
   public String[] getTrustedFolders() {
-    return trustedFolders;
+    return Arrays.copyOf(trustedFolders, trustedFolders.length);
```
bastiandoetsch, line 9: clone()?

bastiandoetsch, line 9: clone()

plugin/src/main/java/io/snyk/languageserver/protocolextension/messageObjects/scanResults/Issue.java
```diff
@@ -4,20 +4,23 @@
...
 		boolean isNew, String filterableIssueType, IgnoreDetails ignoreDetails, AdditionalData additionalData) {
+
 	public String getDisplayTitle() {
 		if (title == null || title.isEmpty()) {
 			return additionalData != null ? additionalData.message() : null;
 		}
-		String displayTitle = title;
+
+		StringBuilder displayTitle = new StringBuilder(title);
+
```
bastiandoetsch, line 14: This is completely unnecessary - the Java compiler takes care of this. It was different in early JDKs, but nowadays there shouldn't be a difference. I'd vote for readability and deactivate the linting rule.

plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigs.java
```diff
@@ -76,7 +76,7 @@ public FolderConfigsParam updateFolderConfigs() {
 
 		for (var project : openProjects) {
 			Path path = ResourceUtils.getFullPath(project);
-			IScopeContext projectScope = new ProjectScope(project);
+			IScopeContext projectScope = new ProjectScope(project); //NOPMD
```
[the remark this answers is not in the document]
acke: Lint doesn't want new objects to be created in the loops, but rather that they are created before the loop, and then reused in the loop. 

However in our case, I do believe we want to create new ProjectScopes here. 

Is there a better solution?

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/filters/BaseFilter.java
acke: The linter complains that the class does not have abstract functions. I solution for not instantiating it was to make the BaseFilter constructor protected, instead of public. (Line 15)

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
acke: Yes, agreed! Ill stick with clone()

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
acke: trustedFolders is a string array and not using complex structures. So we should be good with clone() here?

plugin/src/main/java/io/snyk/languageserver/download/HttpClientFactory.java
acke: The linter had a problem with this construct

if (data.getHost() == null)
    continue;
  return data;
}

And suggested the simplified version
if (data.getHost() != null)
  return data;
}

Removing the continue; and instead go for the not equals was the preferred linter solution.

plugin/src/main/java/io/snyk/languageserver/protocolextension/messageObjects/PublishDiagnostics316Param.java
acke: So we should check for null before cloning?

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/filters/BaseFilter.java
acke: Revert and consider to deactivate rule.

PR #260, opened 2025-02-24
feat: select treenode on snyk showdocument [IDE-957]

acke:
### Description

Will select the treenode when a code lens or a code action is triggered. 
Will display the issue in details view.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs


tests/src/test/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClientTest.java
```diff
@@ -584,4 +585,40 @@ void testGetTotal() throws IOException {
...
+	}
+
+	@Test
+	public void parseQueryString_Returns_Parameters_In_Query_String() throws Exception {
+		cut = new SnykExtendedLanguageClient();
+		String query = "product=Snyk+Code&issueId=7642f506c568056a7090d3ceb7b3c2e0&action=showInDetailPanel";
+		URI uriWithQuery = new URI("snyk://path/to/resource?" + query);
+
+		var result = cut.parseQueryString(uriWithQuery.getQuery());
+		assertEquals(3, result.size());
```
andrewrobinsonhodges-snyk, line 612: I'm not sure that this is adding much value as a separate test. If we put this assert in the test above, we ensure that we're using the same query in both cases, which is more maintainable.

tests/src/test/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClientTest.java
```diff
@@ -584,4 +585,40 @@ void testGetTotal() throws IOException {
...
+	}
+
+	@Test
+	public void parseQueryString_Returns_Empty_Map_If_Query_String_Is_Empty() throws URISyntaxException {
+		cut = new SnykExtendedLanguageClient();
+		URI uriWithoutQuery = new URI("snyk://path/to/resource");
```
andrewrobinsonhodges-snyk, line 618: Why construct the URI if we just pass in null below?

tests/src/test/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClientTest.java
andrewrobinsonhodges-snyk: Alternatively, we could keep this as a separate test, but we should check that the result is of the correct form.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -384,16 +392,111 @@ public void snykScan(SnykScanParam param) {
...
+	}
+
+	public static Map<String, String> parseQueryString(String queryString) {
+		Map<String, String> paramMap = new HashMap<>();
+
+		if (queryString == null || queryString.isEmpty()) {
+			return paramMap;
+		}
+
+		int keyValueSize = 2;
```
andrewrobinsonhodges-snyk, line 484: nitpick, but I don't think this needs to be an explicit variable; we'd never change it, so we could just use it inline.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -384,16 +392,111 @@ public void snykScan(SnykScanParam param) {
...
+		for (String param : queryString.split("&")) {
+			if (!param.isEmpty()) {
+				String[] keyValue = param.split("=");
+
+				if (keyValue.length == keyValueSize) {
```
andrewrobinsonhodges-snyk, line 489: The functionality differs from IntelliJ here. In IntelliJ, we return an empty map if any of the parameters is unable to be parsed. Here we'll ignore the malformed parameter but return the others. We should probably be consisten.

andrewrobinsonhodges-snyk (commented):
Overall function/design is good, but you have added a few methods that could do with renaming for clarity, plus a couple of potential bugs with null checks and hard coded parameters.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykUriUtils.java
```diff
@@ -0,0 +1,55 @@
...
+
+		if (queryString == null || queryString.isEmpty()) {
+			return paramMap;
+		}
+
+		for (String param : queryString.split("&")) {
+			if (!param.isEmpty()) {
+				String[] keyValue = param.split("=");
+
+				if (keyValue.length == 2) { // TODO add the no pmd here
```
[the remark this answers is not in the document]
acke: Will add a ` // NOPMD, AvoidLiteralsInIfCondition` here.

andrewrobinsonhodges-snyk (approved):
I agree that this looks much better. We might want to rename SnykUriDetails or add some explanatory comments later on, as this is pretty specific to this particular URI format at the moment, but it's fine for now.

PR #263, opened 2025-03-05
fix: make sure all images created are disposed

acke:
### Description

- Using Eclipse ImageRegistry to cache images, and make sure they are disposed of correctly.
- fixed preference defaults and migration from secure storage to normal storage
- bumped Eclipse version to 3.1.0

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs

No handle problem with >60000 issues.
image

### DIFF
--- a/plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BaseTreeNode.java
+++ b/plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BaseTreeNode.java
@@ -65,7 +65,7 @@ public class BaseTreeNode extends TreeNode {
 
 	protected ImageDescriptor getImageDescriptor(IResource object) {
 		ILabelProvider labelProvider = null;
-		Image image = null;
+		Image image;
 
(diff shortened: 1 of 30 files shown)


plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BaseTreeNode.java
```diff
@@ -58,16 +60,34 @@ public void removeChildren() {
...
 		} finally {
+			disposeResources(labelProvider, image);
+		}
+	}
+
+	private void disposeResources(ILabelProvider labelProvider, Image image) {
+		if (image != null && !image.isDisposed()) {
+			image.dispose();
+		}
+		if (labelProvider != null) {
```
bastiandoetsch, line 90: I don't think that this is enough. Most likely, we'll also need to cache imageDescriptors for commonly used pics (e.g. severity icons) and renew them once they are disposed.

PR #265, opened 2025-03-07
fix: open/ignored issue filtering [IDE-942]

rrama:
### Description

The previous logic was flipped and had bugs. This PR aims to help improve the understandability of the code by flipping the checking logic, so the predicate logic is what you would expect it to be.

### Checklist

  - No new tests added, as this logic will soon move to the LS.
- [x] Linted
- [x] CHANGELOG.md updated
  - N/A

### Screenshots / GIFs


plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/filters/IgnoresIgnoredIssuesFilter.java
```diff
@@ -8,10 +8,10 @@
 import io.snyk.languageserver.protocolextension.messageObjects.scanResults.Issue;
 
 public class IgnoresIgnoredIssuesFilter extends BaseFilter {
-	private static final Predicate<Issue> predicate = issue -> issue.isIgnored();
+	private static final Predicate<Issue> predicate = issue -> !issue.isIgnored();
```
bastiandoetsch, line 11: Why the inversion? Shouldn't this filter return the ignored issues, not the not-ignored issues?

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/filters/IgnoresOpenIssuesFilter.java
```diff
@@ -1,16 +1,16 @@
...
 
 import java.util.function.Predicate;
 
 import io.snyk.eclipse.plugin.views.snyktoolview.TreeFilterManager;
 import io.snyk.languageserver.protocolextension.messageObjects.scanResults.Issue;
 
-public class IgnoresOpenIssuesFilter extends BaseFilter {	
-	private static final Predicate<Issue> predicate = issue -> !issue.isIgnored();
+public class IgnoresOpenIssuesFilter extends BaseFilter {
+	private static final Predicate<Issue> predicate = issue -> issue.isIgnored();
```
bastiandoetsch, line 11: Same here - IgnoresOpenIssues should return the issues that are not ignored - so why the inversion?

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/filters/IgnoresOpenIssuesFilter.java
```diff
@@ -1,16 +1,16 @@
...
 	
 	public IgnoresOpenIssuesFilter(TreeFilterManager tfm) {
-		super(FILTER_IGNORES_SHOW_IGNORED_ISSUES, predicate, tfm);
+		super(FILTER_IGNORES_SHOW_OPEN_ISSUES, predicate, tfm);
```
bastiandoetsch, line 14: good spot

plugin/src/main/java/io/snyk/languageserver/SnykIssueCache.java
```diff
@@ -56,8 +56,8 @@ public void clearAll() {
 	 * @param path The file path for which issues should be removed
 	 */
 	public void removeAllIssuesForPath(String path) {
-		codeSecurityIssues.clear();
-		codeQualityIssues.clear();
+		codeSecurityIssues.remove(path);
+		codeQualityIssues.remove(path);
```
bastiandoetsch, line 60: Oh wow.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/filters/IgnoresIgnoredIssuesFilter.java
rrama: If you look when the filters are added & removed and then how they are used, it is unintuitively the opposite of what you think. I did also test to ensure that it was correct to flip them.

rrama: Discussed over Tuple. We decided to flip the whole filter predicate logic, to make it easier to reason about.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/filters/IgnoresOpenIssuesFilter.java
rrama: Discussed over Tuple. We decided to flip the whole filter predicate logic, to make it easier to reason about.

PR #273, opened 2025-03-17
fix: sanitize error message, escape characters, add nonce, render as innerText

DariusZdroba:
The error messages rendered in Eclipse could potentially contain HTML or scripts which could be rendered as part of the panel ( no resources were accessible through the scripts but still should not happen )

To fix this, 3 measures have been applied: 

- Use a `Content-Security-Policy` in the `<meta>` tags in the headers, only scripts which have the randomly set `nonce` will run.
- Escape quotes, slashes, greater/less than signs... and unicode characters.
- Render the injected messages as `innerText` - this ensures the message is interpreted as text form and is not executed.
image

### Description

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs


bastiandoetsch (approved):

bastiandoetsch:
@DariusZdroba , please fix the linter stuff and then we'll merge it in :)

PR #297, opened 2025-05-20
feat: added pat to preferences page

acke:
### Description

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs


plugin/src/main/java/io/snyk/eclipse/plugin/preferences/AuthConstants.java
```diff
@@ -0,0 +1,19 @@
...
+    public static final String AUTH_OAUTH2 = "oauth" ;
+    public static final String AUTH_PERSONAL_ACCESS_TOKEN = "pat" ;
+    public static final String AUTH_API_TOKEN = "token";
+    
+    
+    public static final String[][] AUTHENTICATION_OPTIONS = new String[][] {
+        new String[] { "OAuth2", AUTH_OAUTH2 },
+        new String[] { "Personal Access Token", AUTH_PERSONAL_ACCESS_TOKEN },
+        new String[] { "API Token", AUTH_API_TOKEN }
+    };
```
bastiandoetsch, line 14: could we please use something like Pair or an explicit type instead of a 2-dimensional array?

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
```diff
@@ -48,7 +51,6 @@ public class Preferences {
 	public static final String ADDITIONAL_ENVIRONMENT = "ADDITIONAL_ENVIRONMENT";
 	public static final String SEND_ERROR_REPORTS = "SEND_ERROR_REPORTS";
 	public static final String LSP_VERSION = "LSP_VERSION";
-	public static final String USE_TOKEN_AUTH = "useTokenAuth";
```
bastiandoetsch, line 51: this may lead to a changing preference. can we migrate this?

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
```diff
@@ -158,11 +159,11 @@ private void migratePreferences() {
...
-				SEND_ERROR_REPORTS, LSP_VERSION, USE_TOKEN_AUTH, ANALYTICS_PLUGIN_INSTALLED_SENT, ENABLE_DELTA,
-				FILTER_SHOW_CRITICAL, FILTER_SHOW_HIGH, FILTER_SHOW_MEDIUM, FILTER_SHOW_LOW,
-				FILTER_IGNORES_SHOW_OPEN_ISSUES, FILTER_IGNORES_SHOW_IGNORED_ISSUES, FILTER_SHOW_ONLY_FIXABLE,
-				IS_GLOBAL_IGNORES_FEATURE_ENABLED, ENABLE_TELEMETRY, MANAGE_BINARIES_AUTOMATICALLY, ORGANIZATION_KEY,
-				SCANNING_MODE_AUTOMATIC, DEFAULT_ENDPOINT, DEVICE_ID, RELEASE_CHANNEL));
+				SEND_ERROR_REPORTS, LSP_VERSION, ANALYTICS_PLUGIN_INSTALLED_SENT, ENABLE_DELTA, FILTER_SHOW_CRITICAL,
+				FILTER_SHOW_HIGH, FILTER_SHOW_MEDIUM, FILTER_SHOW_LOW, FILTER_IGNORES_SHOW_OPEN_ISSUES,
+				FILTER_IGNORES_SHOW_IGNORED_ISSUES, FILTER_SHOW_ONLY_FIXABLE, IS_GLOBAL_IGNORES_FEATURE_ENABLED,
+				ENABLE_TELEMETRY, MANAGE_BINARIES_AUTOMATICALLY, ORGANIZATION_KEY, SCANNING_MODE_AUTOMATIC,
+				DEFAULT_ENDPOINT, DEVICE_ID, RELEASE_CHANNEL));
```
bastiandoetsch, line 166: this is used to migrate from secure storage to non-secure storage. changing this may fail migration of an old plugin to the newest version.

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
```diff
@@ -328,4 +330,14 @@ public final boolean isSecureStorageReady() {
 	public void setSecureStorageReady(boolean b) {
 		this.secureStorageReady = b;
 	}
+
+	public boolean isAuthenticated() {
+		// TODO check AUTHENTICATION_METHOD, USE_TOKEN_AUTH and AUTH_TOKEN_KEY
+
+		if (getAuthToken().isBlank()) {
```
bastiandoetsch, line 337: is this handling null?

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/PreferencesPage.java
```diff
@@ -24,6 +28,10 @@ public class PreferencesPage extends FieldEditorPreferencePage implements IWorkb
 	private BooleanFieldEditor snykCodeSecurityCheckbox;
 	private BooleanFieldEditor snykCodeQualityCheckbox;
 
+	private static final String PAT_AUTHENTICATION_URL = "https://app.snyk.io/account/personal-access-tokens";
```
bastiandoetsch, line 31: why is this hard-coded? shouldn't it be endpoint specific if it's needed here? and why is it needed to be known here? i would assume the login command to open the browser window to the right URL?

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/PreferencesPage.java
```diff
@@ -39,19 +47,10 @@ protected void createFieldEditors() {
...
-		TokenFieldEditor tokenField = new TokenFieldEditor(prefs, Preferences.AUTH_TOKEN_KEY, "Token:",
-				getFieldEditorParent());
-
-		final var useTokenAuth = new BooleanFieldEditor(Preferences.USE_TOKEN_AUTH,
-				"Use token authentication. It is recommended to keep this turned off, as the default OAuth2 authentication is more secure.",
-				getFieldEditorParent());
-		addField(useTokenAuth);
-		addField(tokenField);
-
-		addField(new StringFieldEditor(Preferences.PATH_KEY, "Path:", 80, getFieldEditorParent()));
```
bastiandoetsch, line 52: why is this removed?

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/PreferencesPage.java
```diff
@@ -88,17 +113,18 @@ protected void createFieldEditors() {
...
 		addField(new LabelFieldEditor("Advanced options:", getFieldEditorParent()));
-		addField(new StringFieldEditor(Preferences.ORGANIZATION_KEY, "Organization:", 80, getFieldEditorParent()));
-		addField(new StringFieldEditor(Preferences.ADDITIONAL_PARAMETERS, "Additional Parameters:", 80,
+		addField(new StringFieldEditor(Preferences.ORGANIZATION_KEY, "Organization:", WIDTH, getFieldEditorParent()));
+		addField(new StringFieldEditor(Preferences.ADDITIONAL_PARAMETERS, "Additional Parameters:", WIDTH,
 				getFieldEditorParent()));
-		addField(new StringFieldEditor(Preferences.ADDITIONAL_ENVIRONMENT, "Additional Environment:", 80,
+		addField(new StringFieldEditor(Preferences.ADDITIONAL_ENVIRONMENT, "Additional Environment:", WIDTH,
 				getFieldEditorParent()));
+		addField(new StringFieldEditor(Preferences.PATH_KEY, "Path:", WIDTH, getFieldEditorParent()));
```
bastiandoetsch, line 121: makes sense, but is it confirmed by @andygongea ?

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/PreferencesPage.java
```diff
@@ -124,11 +150,30 @@ protected void createFieldEditors() {
...
+				performOk(); // This will perform an apply to the values in the PreferencePage and store
+								// these to the PreferenceStore.
+
+				String selectedAuth = getPreferenceStore().getString(Preferences.AUTHENTICATION_METHOD);
+				if (AuthConstants.AUTH_PERSONAL_ACCESS_TOKEN.equals(selectedAuth)) {
+					boolean launched = Program.launch(PAT_AUTHENTICATION_URL);
+					if (!launched) {
+						MessageDialog.openError(getShell(), "Open URL failed",
+								"Failed to launch PAT authentication URL.");
+					}
```
bastiandoetsch, line 171: this should work like IntelliJ & Visual Studio - it should just trigger the login command, which would do the rest. No logic in the IDE for this is needed.

bastiandoetsch: OAuth2 is not handled, yet, right? Just calling the `login` command should be enough and cover it, too.

tests/src/test/java/io/snyk/eclipse/plugin/properties/preferences/PreferencesTest.java
```diff
@@ -71,7 +70,6 @@ void test_DefaultPreferences() {
 		assertEquals("https://api.snyk.io", preferences.getPref(ENDPOINT_KEY));
 		assertEquals("https://downloads.snyk.io", preferences.getPref(CLI_BASE_URL));
 		assertEquals("true", preferences.getPref(SCANNING_MODE_AUTOMATIC));
-		assertEquals("false", preferences.getPref(USE_TOKEN_AUTH));
```
bastiandoetsch, line 74: do we have a test now to test the default for authentication method?

tests/src/test/java/io/snyk/languageserver/LsConfigurationUpdaterTest.java
```diff
@@ -79,7 +78,7 @@ void testGetSettings() {
 			assertEquals("/usr/local/bin/snyk", settings.getCliPath());
 			assertEquals("my-token", settings.getToken());
 			assertEquals("automatic", settings.getScanningMode());
-			assertEquals("token", settings.getAuthenticationMethod());
+			assertEquals("oauth", settings.getAuthenticationMethod());
```
bastiandoetsch, line 81: why not use the constant here?

tests/src/test/java/io/snyk/languageserver/LsConfigurationUpdaterTest.java
```diff
@@ -114,8 +113,8 @@ private void setupPreferenceMock() {
 				.thenReturn("false");
 		when(preferenceMock.getPref(Preferences.CLI_PATH, "")).thenReturn("/usr/local/bin/snyk");
 		when(preferenceMock.getPref(Preferences.AUTH_TOKEN_KEY, "")).thenReturn("my-token");
+		when(preferenceMock.getPref(Preferences.AUTHENTICATION_METHOD, "oauth")).thenReturn("oauth");
```
bastiandoetsch, line 116: same here, I think we have a constant defined for `oauth`

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/PreferencesPage.java
acke: Its moved to Advanced options in the preference page. 

It seemed to be more related to configuring scans, then authentication?

image

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/PreferencesPage.java
acke: The design was
_The “Connect IDE to Snyk“ button opens the PAT generation page (e.g https://app.snyk.io/account/personal-access-tokens) when PAT is the auth method_

But a few questions 
1. Are the pat generation done in each domain? mt-eu, mt-au, and in each single tenant environment?

2. Should we trigger the login workflow, instead of letting the updateConfiguration() when users changes authentication method?

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/PreferencesPage.java
acke: Checking with Andy!

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/AuthConstants.java
acke: Its possible, but the AUTHENTICATION_OPTIONS are used as input to the ComboFieldEditor, so I thought it would make most sense to have the values as `String[][]`.

```
org.eclipse.jface.preference.ComboFieldEditor.ComboFieldEditor(
    String name,
    String labelText,
    String[][] entryNamesAndValues,
    Composite parent
)
```

acke: Yes!

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/PreferencesPage.java
bastiandoetsch: I think this should get a UX review -> e.g. I'd also change the button to `Authenticate...`, but let's verify.

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/AuthConstants.java
```diff
@@ -0,0 +1,27 @@
...
+
+    private static final List<AuthOptionData> OPTION_DEFINITIONS = List.of(
+        new AuthOptionData("OAuth2", AUTH_OAUTH2),
+        new AuthOptionData("Personal Access Token", AUTH_PERSONAL_ACCESS_TOKEN),
+        new AuthOptionData("API Token", AUTH_API_TOKEN)
+    );
+
+    public static final String[][] AUTHENTICATION_OPTIONS = OPTION_DEFINITIONS.stream()
+            .map(optionData -> new String[]{optionData.displayName(), optionData.value()})
+            .toArray(String[][]::new);
```
bastiandoetsch, line 21: I still think that a 2-dimensional array is not a good choice here. It should be a type.

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
```diff
@@ -328,4 +332,14 @@ public final boolean isSecureStorageReady() {
 	public void setSecureStorageReady(boolean b) {
 		this.secureStorageReady = b;
 	}
+
+	public boolean isAuthenticated() {
+		// TODO check AUTHENTICATION_METHOD, USE_TOKEN_AUTH and AUTH_TOKEN_KEY
```
bastiandoetsch, line 337: why?

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/AuthConstants.java
acke: The ComboFieldEditor wants a String[][] entryNamesAndValues, so I do need to convert it to a 2D array somewhere, or what am I missing?

`org.eclipse.jface.preference.ComboFieldEditor.ComboFieldEditor( String name, String labelText, String[][] entryNamesAndValues, Composite parent )`

like this
```
authenticationEditor = new ComboFieldEditor(Preferences.AUTHENTICATION_METHOD, "Authentication Method:",
				AuthConstants.AUTHENTICATION_OPTIONS, getFieldEditorParent());
```

So is there

PR #373, opened 2026-03-30
feat: implement $/snyk.configuration notification handler [IDE-1652]

nick-y-snyk:
## Summary

- Implement \`$/snyk.configuration\` JSON-RPC notification handler for centralized enterprise policy enforcement, replacing the typed \`FolderConfig\`/\`$/snyk.folderConfigs\` infrastructure with a generic \`Map<String, ConfigSetting>\` settings system
- Add durable persistence for global settings, in-memory storage for per-folder configs, change tracking via \`explicitChanges\` set, and outbound \`workspace/didChangeConfiguration\` using \`LspConfigurationParam\` format
- Migrate all consumers (HTMLSettingsPreferencePage, NativeProjectPropertyPage, ReferenceChooserDialog, SnykToolView, LsConfigurationUpdater) to the new \`FolderConfigSettings\` service
- Remove old infrastructure (\`FolderConfig\`, \`FolderConfigsParam\`, \`FolderConfigs\`, \`Settings\`, \`$/snyk.folderConfigs\` handler) and bump LS protocol version to 25
- Align settings registry pattern with VSCode: eliminate inline special-cases, make all three data paths (IDE→LS, LS→IDE, form→IDE) driven by pure registry loops

## What this enables

Enterprise administrators configure security policies via LDX-Sync at tenant/org/machine scope. The Language Server resolves effective settings (applying precedence rules) and pushes them to the Eclipse plugin via \`$/snyk.configuration\`. The plugin persists settings durably, tracks user overrides, and sends changes back via \`workspace/didChangeConfiguration\`. This aligns Eclipse with the IntelliJ implementation for cross-IDE configuration parity.

## Key design decisions

- **Generic map-based settings** (\`Map<String, ConfigSetting>\`) instead of typed fields — any future LS setting works without Eclipse-side code changes
- **Flag-at-modification** change tracking — no baseline diffing; user changes are marked immediately via \`markExplicitlyChanged()\`
- **Full-state outbound** — every \`didChangeConfiguration\` sends all settings with \`changed\` flags, not deltas
- **Folder configs are in-memory only** — LS is authoritative and re-sends on reconnect; incoming config replaces existing entirely (no user-override merge)
- **Lock metadata stored but not yet enforced in UI** — \`isLocked\`, \`source\`, \`originScope\` are persisted for a follow-up iteration
- **Single registry, three paths** — \`LsSettingsRegistry.ENTRIES\` drives outbound serialization, inbound deserialization, and form deserialization; no inline special-cases remain

## Settings registry design (VSCode-aligned)

Each \`LsSettingsRegistry.Entry\` carries:
- \`outboundSerializer\` — pref string → LS wire value (IDE→LS)
- \`inboundDeserializer\` — LS wire value → pref string (LS→IDE, \`$/snyk.configuration\`)
- \`formDeserializer\` — JSON node → pref string (HTML form save; null = use raw \`nodeToString\`)
- \`useInFallbackForm\` — whether the key is included when the fallback settings HTML is active
- \`additionalChangedPrefKeys\` — extra pref keys whose changed-state is OR-ed into the outbound \`changed\` flag (used by severity filters sharing a single changed signal)

All three data paths loop \`ENTRIES\` generically; previously inline-handled keys (\`RISK_SCORE_THRESHOLD\`, \`SEVERITY_FILTER_*\`, \`TRUSTED_FOLDERS\`, \`SCANNING_MODE\`) are now fully registry-driven.

### New files
- \`ConfigSetting.java\` — per-setting value with lock metadata and \`outbound()\` factory
- \`LspFolderConfig.java\` — per-folder config with immutable \`withSetting()\` builder
- \`LspConfigurationParam.java\` — top-level notification/outbound payload type
- \`FolderConfigSettings.java\` — singleton service replacing \`FolderConfigs\` with thread-safe storage, typed convenience methods, and atomic \`computeFolderConfig()\`
- \`LsSettingsKeys.java\` / \`LsFolderSettingsKeys.java\` — LS protocol key constants

### Modified files
- \`SnykExtendedLanguageClient.java\` — added \`snykConfiguration()\` handler, removed old \`folderConfig()\` handler
- \`LsConfigurationUpdater.java\` — single generic loop; removed inline blocks for risk threshold, severity filters, trusted folders
- \`LsSettingsRegistry.java\` — \`RISK_SCORE_THRESHOLD\`, \`SEVERITY_FILTER_*\`, \`TRUSTED_FOLDERS\` moved into \`ENTRIES\`; added \`additionalChangedPrefKeys\` and \`formDeserializer\` fields; \`BY_LS_KEY\` now auto-populated from \`ENTRIES\`
- \`HTMLSettingsPreferencePage.java\` — removed \`FALLBACK_FORM_KEYS\` set and all post-loop special-cases; pure registry loop
- \`FolderConfigSettings.java\` — removed \`mergeConfigs\`; LS values replace existing on \`addAll()\`
- \`Preferences.java\` — added \`explicitChanges\` tracking (mark/is/clear, persisted)
- \`LsBinaries.java\` — \`REQUIRED_LS_PROTOCOL_VERSION\` bumped from \"23\" to \"25\"
- All consumer classes migrated to \`FolderConfigSettings\`

### Deleted files
- \`FolderConfig.java\`, \`FolderConfigsParam.java\`, \`FolderConfigs.java\`, \`Settings.java\`, \`FolderConfigTest.java\`

## Test plan

- [ ] 60 new unit tests added (245 total, all passing, zero regressions)
- [ ] Type deserialization: Gson round-trip for all new types including null fields, various value types, snake_case mapping
- [ ] Notification handler: all payload combinations (global-only, folder-only, both, empty), state reconciliation, scanning mode conversion, exception safety
- [ ] Change tracking: \`explicitChanges\` persistence across restart, mark/is/clear operations, multi-key support
- [ ] Outbound payload: \`LspConfigurationParam\` structure, changed flags from \`explicitChanges\`, no lock metadata in outbound, full-state semantics, reset-to-default (\`value: null, changed: true\`)
- [ ] FolderConfigSettings: singleton, path normalization, typed convenience methods, \`computeFolderConfig\` atomicity, \`isConfigured\`, \`updateFolderConfig\`
- [ ] Consumer migration verified via grep: zero remaining references to old types in production code
- [ ] PMD static analysis passes


plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -409,22 +410,70 @@ public void snykScan(SnykScanParam param) {
...
+				return;
+			}
+
+			if (param.getSettings() != null) {
+				persistGlobalSettings(param.getSettings());
+			}
+
+			if (param.getFolderConfigs() != null) {
+				var folderConfigSettings = FolderConfigSettings.getInstance();
+				folderConfigSettings.addAll(param.getFolderConfigs());
```
bastiandoetsch, line 425: Verification Skill (AI): [Critical] Missing scan trigger and missing `refreshDeltaReference()` after folder configs are received.

The removed `folderConfig()` handler (main branch) did two things this replacement does not:

1. For each folder in the payload, it called `this.triggerScan(path)` when `SCANNING_MODE_AUTOMATIC` was true.
2. After the loop it called `this.toolView.refreshDeltaReference()`.

The new `triggerScan()` method (line 199) guards on `FolderConfigSettings.getAll().isEmpty()`, which returns empty until `snykConfiguration` is received — but nothing calls `triggerScan` after this notification fires. On a fresh startup with automatic scanning enabled, no workspace scan is initiated even though the LS has delivered all required folder metadata.

`refreshDeltaReference()` is still declared in `ISnykToolView` (line 87) and implemented in `SnykToolView` (line 434). Removing it from this code path means the delta baseline is never reset when folder configs are pushed at startup or on LS reconnect.

Test to write first: `snykConfigurationWithFolderConfigs_andAutoScanEnabled_triggersWorkspaceScanAndRefreshesDeltas()` — mock `FolderConfigSettings`, call `snykConfiguration()` with a non-empty `folder_configs` payload and `SCANNING_MODE_AUTOMATIC=true`, then verify both `COMMAND_WORKSPACE_SCAN` is executed and `toolView.refreshDeltaReference()` is called.

plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java
```diff
@@ -0,0 +1,188 @@
...
+	public synchronized void addFolderConfig(LspFolderConfig folderConfig) {
+		if (folderConfig.getFolderPath() == null) {
+			return;
+		}
+		String key = normalizePath(folderConfig.getFolderPath());
+		configs.put(key, folderConfig);
+	}
+
+	public synchronized void addAll(List<LspFolderConfig> folderConfigs) {
+		configs.clear();
```
bastiandoetsch: Verification Skill (AI): [Should Fix] `addAll()` assumes the LS always sends a full replacement. If the `$/snyk.configuration` notification is ever sent as a partial update, `configs.clear()` discards configs for all unmentioned folders before only the changed ones are re-added, leaving those folders unconfigured.

The PR description states "LS is source of truth and re-sends on reconnect," implying full-state semantics. However, this invariant is not documented in the code or enforced by the API, and the LS protocol could change independently of Eclipse.

Suggestion: add a Javadoc comment on `addAll()` that explicitly documents the full-replacement contract and names the protocol version (25) that guarantees it. A one-line comment makes the assumption visible in code review and survives future refactors:

```java
/**
 * Replaces the entire folder-config store with the provided list.
 * Callers must guarantee the list is a complete snapshot (not a delta).
 * Protocol v25 $/snyk.configuration always sends full state.
 */
```

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -409,22 +410,70 @@ public void snykScan(SnykScanParam param) {
...
+		var prefs = Preferences.getInstance();
+		for (var entry : settings.entrySet()) {
+			try {
+				String prefKey = LS_TO_PREF_KEY.get(entry.getKey());
+				if (prefKey == null) {
+					continue;
+				}
+				var setting = entry.getValue();
+				if (setting.getValue() != null) {
+					prefs.store(prefKey, convertInboundValue(entry.getKey(), setting.getValue()));
```
bastiandoetsch: Verification Skill (AI): [Should Fix] `persistGlobalSettings()` calls `prefs.store()` without checking `isLocked`. The `isLocked`, `source`, and `originScope` fields are deserialized from the LS payload (Gson populates them) but are never read in this method.

The PR description acknowledges this: "Lock metadata stored but not yet enforced in UI — isLocked, source, originScope are persisted for a follow-up iteration." This is intentional, but the deferral is invisible in the code itself — a reviewer or future author has no signal that the guard was deliberately omitted.

Suggestion: add an explicit comment at the guard site:

```java
// TODO: skip settings where isLocked=true once client-side enforcement is implemented (follow-up ticket)
if (setting.getValue() != null) {
    prefs.store(prefKey, convertInboundValue(entry.getKey(), setting.getValue()));
}
```

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
```diff
@@ -112,13 +112,12 @@ public Object getInitializationOptions(URI rootUri) {
 			waitForInit();
 		}
 
-		Settings currentSettings = null;
 		try {
-			currentSettings = new LsConfigurationUpdater().getCurrentSettings();
+			return new LsConfigurationUpdater().buildConfigurationParam();
```
bastiandoetsch: Verification Skill (AI): [Should Fix] `getInitializationOptions()` catches only `IllegalStateException | IllegalArgumentException`; other runtime exceptions from `buildConfigurationParam()` escape uncaught and abort the LS handshake.

`buildConfigurationParam()` now calls `FolderConfigSettings.getInstance().getAll()`, reads `Activator.PLUGIN_VERSION` (can be `null` before bundle start), constructs `IssueViewOptions` and `FilterSeverity` objects from `Preferences` fields, and invokes `FolderConfigSettings.normalizePath()`. Any `NullPointerException` from a null `Preferences` field, or a `ConcurrentModificationException`, propagates past this catch and prevents the plugin from starting.

The old `getCurrentSettings()` had the same narrow catch, but `buildConfigurationParam()` is substantially larger and calls more code paths.

Test to write first: `getInitializationOptions_withUnexpectedRuntimeException_returnsNullAndDoesNotThrow()` — inject a broken `Preferences` mock that throws `NullPointerException` from `getBooleanPref()` and verify the method returns `null`.

Fix: broaden the catch to `Exception` (already used in `snykConfiguration()` at line 429 of `SnykExtendedLanguageClient`).

bastiandoetsch:
Verification Skill (AI): [Should Fix] **`hasAuthenticated()` stores `AUTH_TOKEN_KEY` via `prefs.store()` instead of `storeAndTrackChange()`** — the token is never marked `isExplicitlyChanged` after an OAuth flow.

**File:** `plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java`, line 335 (`hasAuthenticated` handler — not modified in this PR but relevant to the change).

`buildConfigurationParam()` (LsConfigurationUpdater.java line 57–59) sends the token outbound with `preferences.isExplicitlyChanged(Preferences.AUTH_TOKEN_KEY)` as the `changed` flag. Because `hasAuthenticated()` calls `prefs.store(AUTH_TOKEN_KEY, newToken)` rather than `prefs.markExplicitlyChanged(AUTH_TOKEN_KEY)`, an OAuth-sourced token is always sent back to the LS with `changed=false`, telling the LS the user has not explicitly chosen this token.

`TokenFieldEditor.doStore()` correctly calls `markExplicitlyChanged` for manual token entry, so the gap is specifically the OAuth callback path.

Test to write first: `hasAuthenticated_withNewToken_marksAuthTokenAsExplicitlyChanged()` — call `hasAuthenticated()` with a new token value, then assert `preferences.isExplicitlyChanged(Preferences.AUTH_TOKEN_KEY)` is `true`. Currently fails.

Fix: add `prefs.markExplicitlyChanged(Preferences.AUTH_TOKEN_KEY)` in the `if (differentToken)` block of `hasAuthenticated()`.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: 1. should be handled in ls
2. done

bastiandoetsch: Verification Skill (AI): Fixed. `triggerScan(null)` is called at line 427 when automatic scanning is enabled, and `refreshDeltaReference()` is called at line 430 — both code paths are now present in `snykConfiguration()`. No further action needed.

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
bastiandoetsch: Verification Skill (AI): This is still present. `getInitializationOptions()` at `plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java` line 123 still catches only `IllegalStateException | IllegalArgumentException`. A `NullPointerException` or `ConcurrentModificationException` from the expanded `buildConfigurationParam()` call will still propagate uncaught and abort the LS handshake.

bastiandoetsch:
Verification Skill (AI): Re finding on `hasAuthenticated()` token tracking: finding is withdrawn. `LsKey.TOKEN` is registered in `LsSettingsRegistry` (line 114) with `Entry.alwaysChanged`, which unconditionally sets `changed=true` in `buildConfigurationParam()`. The `isExplicitlyChanged` path is never reached for the token. I should have checked the registry entry before flagging this.

plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java
bastiandoetsch: Verification Skill (AI): This is still present. `addAll()` at `plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java` line 43 still has no Javadoc documenting the full-replacement contract. The inline comment on line 44 mentions removing stale folders but does not state that callers must guarantee a complete snapshot.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
bastiandoetsch: Verification Skill (AI): This is still present. `persistGlobalSettings()` at `plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java` line 447 still does not check `isLocked`, and there is no TODO comment indicating the deferral was intentional.

tests/src/test/java/io/snyk/languageserver/LsConfigurationUpdaterTest.java
```diff
@@ -2,91 +2,440 @@
...
+	void testBuildConfigurationParam_changedFlagFromExplicitChanges() {
+		setupPreferenceMock();
+		when(preferenceMock.isExplicitlyChanged(Preferences.ENDPOINT_KEY)).thenReturn(true);
+		when(preferenceMock.isExplicitlyChanged(Preferences.ORGANIZATION_KEY)).thenReturn(false);
+
+		var param = new LsConfigurationUpdater().buildConfigurationParam();
+		Map<String, ConfigSetting> settings = param.getSettings();
+
+		assertTrue(settings.get(LsKey.ENDPOINT.key).getChanged());
+		assertFalse(settings.get(LsKey.ORGANIZATION.key).getChanged());
```
bastiandoetsch, line 71: Verification Skill (AI): [Critical] Three tests fail on CI (Java 17) with `AssertionFailedError: expected: <false> but was: <true>`. The three failing tests are:

1. `testBuildConfigurationParam_changedFlagFromExplicitChanges` (line 71) — asserts `organization.changed == false` when `isExplicitlyChanged(ORGANIZATION_KEY) = false`, but `buildConfigurationParam()` also marks changed when the value differs from `outboundDefault`. The mock returns `"organization"` for `ORGANIZATION_KEY` with default `""`, so `!outboundDefault.equals(rawValue)` is true → `changed = true`.

2. `testBuildConfigurationParam_issueViewIgnoredIssuesUnchangedWhenNotExplicitlyChanged` (line 346) — no `getPref(FILTER_IGNORES_SHOW_IGNORED_ISSUES, "false")` stub configured in `setupPreferenceMock()`. Mockito returns `null` → `outboundDefault.equals(null)` is false → `changed = true`.

3. `testBuildConfigurationParam_severityFilterUnchangedWhenNoSeverityChanged` (line 385) — the mock returns `getPref(FILTER_SHOW_HIGH, "true") = "false"`. The default is `"true"`, so `!"true".equals("false")` = true → `SEVERITY_FILTER_HIGH.changed = true`.

Root cause: `buildConfigurationParam()` uses a dual-trigger for `changed`: explicit tracking OR deviation from `outboundDefault`. The `testBuildConfigurationParam_nonDefaultEndpointSendsChangedAfterRestart` test validates that the non-default fallback is intentional (post-restart behavior). The failing tests contradict this by expecting `changed = false` for non-default values that are not explicitly changed.

The fix must reconcile these two semantics. Either:
- The `setupPreferenceMock()` must stub all `getPref` calls to return their default values for settings not under test; OR
- The non-default fallback must be removed from `buildConfigurationParam()` and replaced by persistence of explicit changes across restart (the `explicitChanges` set already serves this role if populated before the first `buildConfigurationParam()` call).

Test to write first: add stubs for all `getPref` calls that return values matching `outboundDefault` in `setupPreferenceMock()`, then run the suite.

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
```diff
@@ -374,4 +388,47 @@ public boolean isAuthenticated() {
...
+	public void clearExplicitlyChanged(String key) {
+		if (key == null) {
+			return;
+		}
+		explicitChanges.remove(key);
+		persistExplicitChanges();
+	}
+
+	private void persistExplicitChanges() {
+		insecurePreferences.put(EXPLICIT_CHANGES_KEY, String.join(",", explicitChanges));
```
bastiandoetsch, line 437: Verification Skill (AI): [Should Fix] Comma-separated serialisation of `explicitChanges` is fragile. This is still present: `String.join(",", explicitChanges)` at line 427 splits on `","` at construction (line 156-162) with no escaping. All current `Preferences.*` and `LsKey.*` values are comma-free today, but the design goal of accommodating future LS keys without Eclipse changes makes a comma-containing key realistic. Use `"|"`, `"\n"`, or a JSON array instead. Alternatively, URL-encode each key before joining.

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
```diff
@@ -112,13 +114,16 @@ public Object getInitializationOptions(URI rootUri) {
...
 
-		Settings currentSettings = null;
 		try {
-			currentSettings = new LsConfigurationUpdater().getCurrentSettings();
+			var param = new LsConfigurationUpdater().buildConfigurationParam();
+			// Pre-serialize to JsonElement so LSP4J embeds the JSON directly
+			// instead of re-serializing through its own Gson instance, which
+			// can drop fields from custom POJOs like ConfigSetting.
+			return new Gson().toJsonTree(param);
 		} catch (IllegalStateException | IllegalArgumentException e) {
```
bastiandoetsch: Verification Skill (AI): [Should Fix] `getInitializationOptions()` still only catches `IllegalStateException | IllegalArgumentException`. `buildConfigurationParam()` now calls `FolderConfigSettings.getInstance().getAll()`, `Activator.PLUGIN_VERSION` (can be null before bundle start), and `LsConfigurationUpdater` which iterates `LsSettingsRegistry.ENTRIES`. Any `NullPointerException` or `ConcurrentModificationException` escapes uncaught and aborts the LS handshake. `snykConfiguration()` in the same class already uses `catch (Exception e)` — apply the same pattern here.

bastiandoetsch (commented):
Verification Skill (AI): Automated verification findings (re-run) — CI build is failing; see inline comments for details.

plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java
bastiandoetsch: Verification Skill (AI): This is still present. `addAll()` at `plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java:43` still silently clears configs for unmentioned folders without a Javadoc comment documenting the full-replacement contract. Add a Javadoc comment naming the protocol version (v25) and the full-state guarantee.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
bastiandoetsch: Verification Skill (AI): This is still present. `persistGlobalSettings()` at `plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java:438` reads `setting.getValue()` without checking `setting.getIsLocked()`, and no TODO comment was added. The deliberate deferral is invisible to future reviewers.

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
bastiandoetsch: Verification Skill (AI): This is still present. `getInitializationOptions()` at `plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java:123` still only catches `IllegalStateException | IllegalArgumentException`. Any `NullPointerException` from `buildConfigurationParam()` escapes uncaught. Broaden to `catch (Exception e)` as `snykConfiguration()` already does.

tests/src/test/java/io/snyk/languageserver/LsConfigurationUpdaterTest.java
```diff
@@ -2,91 +2,440 @@
...
+
+		assertNotNull(settings.get(LsKey.ENDPOINT.key));
+		assertEquals("endpoint", settings.get(LsKey.ENDPOINT.key).getValue());
+
+		assertNotNull(settings.get(LsKey.ORGANIZATION.key));
+		assertEquals("organization", settings.get(LsKey.ORGANIZATION.key).getValue());
+	}
+
+	@Test
+	void testBuildConfigurationParam_changedFlagFromExplicitChanges() {
```
bastiandoetsch, line 62: **[CRITICAL] Three tests fail on CI (Java 17)**

`testBuildConfigurationParam_changedFlagFromExplicitChanges` (line 62), `testBuildConfigurationParam_issueViewIgnoredIssuesUnchangedWhenNotExplicitlyChanged` (line 342), and `testBuildConfigurationParam_severityFilterUnchangedWhenNoSeverityChanged` (line 381) all fail because `setupPreferenceMock()` returns non-default values for settings not under test. The third condition in `buildConfigurationParam()` — `!outboundDefault.equals(rawValue)` — then fires unexpectedly and sets `changed=true`.

Fix: add `getPref` stubs in `setupPreferenceMock()` that return the `outboundDefault` value for every setting not actively under test. Example: add `when(preferenceMock.getPref(Preferences.FILTER_IGNORES_SHOW_IGNORED_ISSUES, "false")).thenReturn("false")` (and similarly for all other unstubbed boolean-pref keys).

plugin/src/main/java/io/snyk/languageserver/LsSettingsRegistry.java
```diff
@@ -0,0 +1,225 @@
...
+                    StringBuilder sb = new StringBuilder();
+                    for (com.fasterxml.jackson.databind.JsonNode el : node) {
+                        if (sb.length() > 0) sb.append(java.io.File.pathSeparator);
+                        sb.append(el.asText());
+                    }
+                    return sb.toString();
+                }));
+        entries.put(LsKey.CLI_RELEASE_CHANNEL,    Entry.fallback(LsKey.CLI_RELEASE_CHANNEL, Preferences.RELEASE_CHANNEL, "stable"));
+
+        ENTRIES = Collections.unmodifiableMap(entries);
```
bastiandoetsch, line 202: **[HIGH] `ADDITIONAL_ENV` and `ENABLE_TELEMETRY` dropped from protocol**

`LsKey.ADDITIONAL_ENV` (`"additional_environment"`) and `LsKey.ENABLE_TELEMETRY` (`"enableTelemetry"`) are defined in `LsKey.java` (lines 14 and 16) but have zero entries in `ENTRIES`. The old `LsConfigurationUpdater.getCurrentSettings()` explicitly included both. After this change, the language server never receives additional environment variables or the telemetry opt-out — silently breaking users who configure proxy via `ADDITIONAL_ENVIRONMENT` (a common enterprise use case).

Fix: add entries to `ENTRIES` for both keys, e.g.:
```java
Entry.simple(LsKey.ADDITIONAL_ENV, Preferences.ADDITIONAL_ENVIRONMENT, "")
Entry.bool(LsKey.ENABLE_TELEMETRY, Preferences.ENABLE_TELEMETRY, false)
```
Add regression tests asserting these keys appear in `buildConfigurationParam().getSettings()`.

plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java
```diff
@@ -0,0 +1,191 @@
...
+	public synchronized void addAll(List<LspFolderConfig> folderConfigs) {
+		// Collect paths present in the incoming list; remove folders no longer reported by LS.
+		Set<String> incomingKeys = new HashSet<>();
+		for (LspFolderConfig incoming : folderConfigs) {
+			if (incoming.getFolderPath() == null) {
+				continue;
+			}
+			String key = normalizePath(incoming.getFolderPath());
+			incomingKeys.add(key);
+			configs.put(key, incoming);
```
bastiandoetsch, line 57: **[HIGH] `addAll()` overwrites user-modified folder configs**

This unconditional `configs.put(key, incoming)` discards any folder config changes the user made via `ReferenceChooserDialog` or `NativeProjectPropertyPage` the moment the LS sends a `$/snyk.configuration` notification (which now happens much more frequently than the old `$/snyk.folderConfigs`).

Scenario: user sets `baseBranch` to `"feature/my-branch"` → `computeFolderConfig()` stores the change → LS sends `$/snyk.configuration` → `addAll()` overwrites → user's change is gone.

Fix: merge incoming LS data with user overrides. Only overwrite settings where `ConfigSetting.getChanged() == false` (i.e., not user-set). User-set settings (`changed=true`) should be preserved until explicitly reset.

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
```diff
@@ -112,13 +114,16 @@ public Object getInitializationOptions(URI rootUri) {
...
 		}
 
-		Settings currentSettings = null;
 		try {
-			currentSettings = new LsConfigurationUpdater().getCurrentSettings();
+			var param = new LsConfigurationUpdater().buildConfigurationParam();
+			// Pre-serialize to JsonElement so LSP4J embeds the JSON directly
+			// instead of re-serializing through its own Gson instance, which
+			// can drop fields from custom POJOs like ConfigSetting.
+			return new Gson().toJsonTree(param);
```
bastiandoetsch, line 122: **[HIGH] Gson/OSGi field reflection risk**

`ConfigSetting` has private fields (`value`, `changed`, `source`, `originScope`, `isLocked`) with only package-private setters. Gson accesses private fields via reflection. In an OSGi environment, cross-bundle field reflection requires explicit `java.lang.reflect` module opens — without them, `InaccessibleObjectException` (Java 17+) causes Gson to serialize each `ConfigSetting` as `{}`, meaning the LS receives initialization options where all settings have `null` value and `null` changed.

Verify: confirm `new Gson().toJsonTree(param)` produces correct JSON in the actual OSGi container (unit tests run outside OSGi and won't catch this). Consider making `ConfigSetting` fields public, adding `@Expose` annotations with `GsonBuilder.excludeFieldsWithoutExposeAnnotation()`, or extracting a `toMap()` helper to avoid reflection entirely.

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
```diff
@@ -24,74 +20,61 @@ public void configurationChanged() {
...
+				io.snyk.languageserver.download.LsBinaries.REQUIRED_LS_PROTOCOL_VERSION);
+		param.setIntegrationName(Activator.INTEGRATION_NAME);
+		param.setIntegrationVersion(Activator.PLUGIN_VERSION);
+		param.setRuntimeName(org.apache.commons.lang3.SystemUtils.JAVA_RUNTIME_NAME);
+		param.setRuntimeVersion(org.apache.commons.lang3.SystemUtils.JAVA_RUNTIME_VERSION);
+		param.setOsArch(org.apache.commons.lang3.SystemUtils.OS_ARCH);
+		param.setOsPlatform(org.apache.commons.lang3.SystemUtils.OS_NAME);
+		param.setPath(preferences.getPref(Preferences.PATH_KEY, ""));
+		param.setDeviceId(preferences.getPref(Preferences.DEVICE_ID, ""));
+		param.setTrustedFolders(trustedFolders);
```
bastiandoetsch, line 74: **[HIGH] `trusted_folders` serialized twice**

`trusted_folders` ends up in the payload both inside `settings["trusted_folders"]` (as a `ConfigSetting`, via the ENTRIES loop) and at the top level as a plain `String[]` via this `setTrustedFolders()` call. Whether intentional (the comment mentions "InitializationOptions") or accidental duplication depends on the LS protocol v25 spec. Either way it's undocumented.

Please add a comment here explaining why `trusted_folders` appears at both levels. If it should only be at one level, remove the duplicate.

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
```diff
@@ -24,74 +20,61 @@ public void configurationChanged() {
...
+			if (LsKey.CLI_PATH == entry.lsKey) {
+				rawValue = preferences.getCliPath();
+			} else {
+				rawValue = preferences.getPref(entry.prefKey, entry.outboundDefault);
+			}
+
+			Object lsValue = entry.outboundSerializer.apply(rawValue);
+
+			boolean changed = entry.isAlwaysChanged || preferences.isExplicitlyChanged(entry.prefKey)
+					|| (entry.outboundDefault != null && !entry.outboundDefault.equals(rawValue));
```
bastiandoetsch: **[MEDIUM] Two conflicting definitions of "changed"**

The third condition (`!entry.outboundDefault.equals(rawValue)`) means any key whose value in the preference store differs from `outboundDefault` — even if the user never touched it — gets `changed=true`. This creates a feedback loop: LS sends config → plugin persists via `persistGlobalSettings()` → next `didChangeConfiguration` sends `changed=true` for those keys → LS interprets them as user-set overrides.

Specific case: `ENDPOINT_KEY` has `outboundDefault=""` in the registry but `insecureStore.setDefault(ENDPOINT_KEY, "https://api.snyk.io")`. On a fresh install `rawValue = "https://api.snyk.io"` (from the store default), which differs from `outboundDefault=""`, so `changed=true` is sent immediately — even though the user has never touched the endpoint.

Fix: align `outboundDefault` in `LsSettingsRegistry` with the actual `insecureStore` defaults in `Preferences.java`. The two default systems must be consistent.

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/PreferencesPage.java
```diff
@@ -222,7 +225,26 @@ public static void notifyAuthTokenChanged(String token) {
...
+		for (String key : trackedKeys) {
+			oldValues.put(key, preferences.getPref(key, ""));
+		}
+
 		boolean superOK = super.performOk();
+
+		for (String key : trackedKeys) {
+			String newValue = preferences.getPref(key, "");
+			if (!Objects.equals(oldValues.get(key), newValue)) {
+				preferences.markExplicitlyChanged(key);
```
bastiandoetsch, line 244: **[MEDIUM] `explicitChanges` never cleared when user resets to default**

`markExplicitlyChanged(key)` is called when value changes, but `clearExplicitlyChanged()` is never called when a value is reset to its default. If a user sets `ENDPOINT_KEY` to a custom URL, saves, then resets it back to the default, `ENDPOINT_KEY` remains in `explicitChanges` permanently — it can never be cleared via the native `PreferencesPage`. Over time the persisted set accumulates stale keys.

Fix: after the old/new value comparison, also call `clearExplicitlyChanged(key)` when `newValue` equals the registered `outboundDefault` for that key.

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
```diff
@@ -374,4 +388,47 @@ public boolean isAuthenticated() {
...
+		try {
+			insecurePreferences.flush();
```
bastiandoetsch, line 439: **[MEDIUM] Synchronous `flush()` on UI thread per-key change**

`persistExplicitChanges()` calls `insecurePreferences.flush()` (a disk I/O operation) synchronously. `markExplicitlyChanged()` is called from `performOk()` for each changed key — up to ~25 `flush()` calls in sequence on the Eclipse UI thread — and also from `BaseHandler.execute()` for filter changes (also UI thread).

Fix: batch the persist — call `persistExplicitChanges()` once at the end of `performOk()` rather than on each individual key change. For handler cases, debounce or async-schedule the flush.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -409,21 +409,47 @@ public void snykScan(SnykScanParam param) {
...
+
+			if (param.getSettings() != null) {
+				persistGlobalSettings(param.getSettings());
+			}
+
+			if (param.getFolderConfigs() != null) {
+				var folderConfigSettings = FolderConfigSettings.getInstance();
+				folderConfigSettings.addAll(param.getFolderConfigs());
+				if (Preferences.getInstance().getBooleanPref(Preferences.SCANNING_MODE_AUTOMATIC)) {
+					triggerScan(null);
```
bastiandoetsch, line 427: **[MEDIUM] `triggerScan(null)` fires workspace-wide scan instead of per-folder**

The old `folderConfig()` handler called `triggerScan(path)` for each newly-ready folder. This `triggerScan(null)` triggers a full workspace scan every time the LS sends any `$/snyk.configuration` notification, even for incremental single-folder updates. On large multi-folder workspaces this causes excessive CPU load and UI noise.

Fix: if `param.getFolderConfigs()` contains the affected folders, iterate them and call `triggerScan(path)` for each — matching the old behavior.

plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java
```diff
@@ -0,0 +1,191 @@
...
+import io.snyk.eclipse.plugin.utils.SnykLogger;
+import io.snyk.languageserver.LsFolderSettingsKeys;
+import io.snyk.languageserver.protocolextension.messageObjects.ConfigSetting;
+import io.snyk.languageserver.protocolextension.messageObjects.LspFolderConfig;
+
+public class FolderConfigSettings {
+
+	private static FolderConfigSettings instance;
+
+	private final Map<String, LspFolderConfig> configs = new ConcurrentHashMap<>();
```
bastiandoetsch, line 22: **[MEDIUM] Inconsistent locking: `ConcurrentHashMap` + `synchronized` methods**

All public methods are `synchronized`, so the instance lock already provides full thread safety for compound operations. The `ConcurrentHashMap` is redundant — it's never accessed concurrently without the synchronized lock (since all callers go through synchronized methods). Two locking strategies running in parallel is confusing.

Fix: replace `ConcurrentHashMap` with a plain `HashMap` (all access is behind `synchronized`). Or drop `synchronized` from read-only methods (e.g., `getAll()`, `isConfigured()`) and document that only compound operations need `synchronized` — but pick one strategy.

plugin/src/main/java/io/snyk/languageserver/LsSettingsRegistry.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    private static final String TRUE = "true";
+
+    /**
+     * Ordered outbound entries keyed by LsKey.
+     * Iterated by LsConfigurationUpdater.buildConfigurationParam().
+     * EnumMap preserves insertion order.
+     */
+    public static final Map<LsKey, Entry> ENTRIES;
```
bastiandoetsch, line 102: **[MEDIUM] `LsSettingsRegistry` is doing too much — 8+ field Entry**

`Entry` has 9 constructor parameters including three functional interface fields. The class serves simultaneously as: outbound serializer, inbound deserializer, form deserializer, changed-flag metadata, fallback-form membership flag, and additional changed-key group. The `SCANNING_MODE` and `RISK_SCORE_THRESHOLD` entries bypass factory methods and use the raw 9-argument constructor, meaning future contributors must understand the full signature.

Suggestion: extract a builder for `Entry` to make parameter intent explicit. Consider separating the inbound deserialization concern from outbound serialization — they serve different callers and the coupling makes the class hard to modify safely.

plugin/src/main/resources/ui/html/settings-fallback.html
```diff
@@ -337,11 +370,11 @@ <h2>CLI Configuration</h2>
...
         isFallbackForm: true,
-        cliPath: get('cliPath').value,
-        manageBinariesAutomatically: get('manageBinariesAutomatically').checked,
-        cliBaseDownloadURL: get('cliBaseDownloadURL').value || get('cliBaseDownloadURL').placeholder,
-        cliReleaseChannel: get('cliReleaseChannel').value,
-        insecure: get('insecure').checked
+        cli_path: get('cli_path').value,
+        automatic_download: get('automatic_download').checked,
+        binary_base_url: get('binary_base_url').value || get('binary_base_url').placeholder,
+        cli_release_channel: get('cli_release_channel').value === 'custom' ? (function() { var v = (get('cli_release_channel_custom').value || '').trim(); if (v && v.charAt(0) !== 'v') { v = 'v' + v; } return v || 'stable'; })() : get('cli_release_channel').value,
```
bastiandoetsch, line 376: **[LOW] Empty custom CLI version silently falls back to `"stable"`**

If the user selects "Specify version…" (`custom`) but leaves the text field empty (or clears it), the sent value is `"stable"` with no user-visible warning. `validateCliVersion()` only validates the format of non-empty strings, not whether a non-empty value was provided when `custom` is selected.

Fix: treat empty custom version as invalid in `validateCliVersion()` when `cli_release_channel === 'custom'`. Block form submission when empty.

bastiandoetsch (commented):
## Adversarial Review — feat/IDE-1652

**Verdict: REJECT** — Two blocking behavioral regressions must be fixed before merge. Additional medium/low findings inline.

### Blocking (must fix)
1. `ADDITIONAL_ENV` and `ENABLE_TELEMETRY` are defined in `LsKey` but missing from `LsSettingsRegistry.ENTRIES` — they will never be sent to LS after this change.
2. `FolderConfigSettings.addAll()` overwrites user-modified folder configs on every LS `$/snyk.configuration` notification.

### What went well
- `LsKey` enum eliminates scattered wire-protocol string literals — a real improvement.
- Strong test coverage for the new subsystems (`FolderConfigSettingsTest`, `LsConfigurationUpdaterTest`, `PreferencesTest`).
- `LspFolderConfig.withSetting()` immutable-update pattern is clean and well-implemented.

tests/src/test/java/io/snyk/languageserver/LsConfigurationUpdaterTest.java
nick-y-snyk: All three tests pass in the current code (27/27, see surefire report). The third condition `!outboundDefault.equals(rawValue)` described in the finding does not exist in `buildConfigurationParam()` — `changed` is set only via `entry.isAlwaysChanged || preferences.isExplicitlyChanged(entry.prefKey)` (plus `additionalChangedPrefKeys` OR). The reviewer appears to have analysed an older commit. No action needed.

plugin/src/main/java/io/snyk/languageserver/LsSettingsRegistry.java
nick-y-snyk: These two settings are not LS protocol keys and were never sent via `didChangeConfiguration` — they are env-var-only. `ADDITIONAL_ENVIRONMENT` is parsed and injected into the LS subprocess environment in `LsRuntimeEnvironment.addAdditionalParamsAndEnv()`. `ENABLE_TELEMETRY` (i.e. `SNYK_CFG_DISABLE_ANALYTICS`) is set as a process env var in `LsRuntimeEnvironment.addTelemetry()`. The canonical telemetry settings key is `SEND_ERROR_REPORTS`, which is present in `ENTRIES`. `LsKey.ADDITIONAL_ENV` and `LsKey.ENABLE_TELEMETRY` appear to be stale enum values — no action needed for ENTRIES.

nick-y-snyk: Addendum: `additional_environment` (the global pref key) vs `additional_env` (the v25 folder-scoped LS key, `LsFolderSettingsKeys.ADDITIONAL_ENV`) are two different things. In protocol v25, `additional_env` is folder-scoped and arrives **inbound** from the LS via `$/snyk.configuration` — the LS owns it per-folder, the IDE does not send it outbound in `didChangeConfiguration`. `LsRuntimeEnvironment` still reads the global pref as a subprocess env-var fallback at startup (before the LS sends folder configs), which is intentional. So the global pref path and the folder-scoped LS protocol key serve different purposes. Still no action needed for ENTRIES.

plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java
nick-y-snyk: This was intentional. `mergeConfigs()` existed in commit `435413a` but was explicitly removed in `db996b9` to align with VSCode behaviour — LS values are authoritative for folder config state. The expected flow is: user saves `base_branch` → IDE calls `configurationChanged()` → LS receives updated value → LS sends back `$/snyk.configuration` with the acknowledged value → `addAll()` stores it. The race window (user save → LS ack) is small and bounded. No change planned.

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
nick-y-snyk: Already addressed. `getInitializationOptions()` pre-serializes `LspConfigurationParam` to `JsonElement` (line 122) specifically because LSP4J's own Gson instance drops `ConfigSetting` fields from custom POJOs — the comment on lines 120–121 documents this. The OSGi reflection concern doesn't apply here: `com.google.gson` is declared as a bundle dependency in `MANIFEST.MF`, Eclipse's OSGi classloader allows private field access for bundles that import the package, and Gson has been used this way in Eclipse plugins (including lsp4j itself) throughout Java 17. All 27 tests pass including the round-trip serialization test. No action needed.

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
nick-y-snyk: Intentional dual placement. `LspConfigurationParam` is used for both `initialize` and `didChangeConfiguration`. The top-level `trustedFolders` field populates `initializationOptions.trustedFolders` (required at handshake). The `settings["trusted_folders"]` entry populates the `settings` map for `didChangeConfiguration`. Both paths are required by the LS protocol. This is documented in `LsSettingsRegistry.java` line 174: `// trusted_folders: always-changed array, also set top-level for InitializationOptions.`

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
nick-y-snyk: The `!outboundDefault.equals(rawValue)` condition no longer exists. `buildConfigurationParam()` line 47 is: `boolean changed = entry.isAlwaysChanged || preferences.isExplicitlyChanged(entry.prefKey);` — only explicit tracking, no value-comparison fallback. The feedback loop and ENDPOINT default mismatch described were consequences of that removed condition. No action needed.

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/PreferencesPage.java
nick-y-snyk: Acknowledged. The native `PreferencesPage` is being dropped in the next PR — this won't be addressed here.

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
nick-y-snyk: Fixed. Added `markExplicitlyChangedNoFlush()`, `clearExplicitlyChangedNoFlush()`, and `flushExplicitChanges()` to `Preferences`. The `HTMLSettingsPreferencePage` ENTRIES loop now calls the no-flush variants and flushes once after the loop. `ExecuteCommandBridge.saveLoginArgs()` does the same — 3 marks then 1 flush. The `PreferencesPage` native path is unchanged (being dropped next PR). Tests pass, PMD clean.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Intentional. The old `$/snyk.folderConfig` protocol notification was inherently per-folder, so per-folder `triggerScan(path)` was correct. The new `$/snyk.configuration` sends a full-state snapshot of all folders in one notification — there is no signal for which folders actually changed. Iterating `param.getFolderConfigs()` and firing N individual folder scans would be worse than one workspace scan. `triggerScan(null)` → `COMMAND_WORKSPACE_SCAN` is the correct semantics for a full-state notification.

plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java
nick-y-snyk: Fixed. Kept `ConcurrentHashMap` for its lock-striping on individual puts/gets; removed `synchronized` from single-operation methods (`addFolderConfig`, `getFolderConfig`, `isConfigured`, `updateFolderConfig`) where `ConcurrentHashMap` provides sufficient thread safety. Kept `synchronized` on compound operations (`addAll` — put loop + retainAll, `getAll` — consistent snapshot, `computeFolderConfig` — read-compute-put) where atomicity requires the instance lock. Tests pass, PMD clean.

plugin/src/main/java/io/snyk/languageserver/LsSettingsRegistry.java
nick-y-snyk: Acknowledged as a design concern. The 9-param constructor is private and only used directly by 4 entries (`SCANNING_MODE`, `RISK_SCORE_THRESHOLD`, `TRUSTED_FOLDERS`, severity filters) that need non-standard serializers. All other entries go through the named factory methods (`simple`, `bool`, `fallback`, etc.) which already make intent explicit. A builder would add ~40 lines of boilerplate without reducing complexity at the 4 raw call sites. Not addressing in this PR.

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
nick-y-snyk: Fixed. Broadened to `catch (Exception e)` to match `snykConfiguration()`'s pattern and prevent any unchecked exception from aborting the LS handshake.

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
nick-y-snyk: Fixed. Broadened to `catch (Exception e)`.

nick-y-snyk: Fixed. Broadened to `catch (Exception e)`.

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
nick-y-snyk: Not changing. All `explicitChanges` entries are pref key constants defined in `Preferences.java` (e.g. `ENDPOINT_KEY`, `ORGANIZATION_KEY`) — commas are structurally impossible in these identifiers. The format is safe as-is.

plugin/src/main/resources/ui/html/settings-fallback.html
nick-y-snyk: Not fixing here — `settings-fallback.html` is served from the LS repo. The fix needs to go there.

plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java
nick-y-snyk: Added Javadoc documenting the full-replacement contract and the LS protocol v25 guarantee.

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
nick-y-snyk: Added inline comment: `isLocked` not checked intentionally — IDE persists LS-pushed values regardless of lock so they survive restarts; lock enforcement is the LS's responsibility on the next send.

plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java
nick-y-snyk: Added Javadoc.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Added inline comment explaining the intentional deferral.

PR #391, opened 2026-05-15
feat: verify CLI protocol version on LS start [IDE-1652]

nick-y-snyk:
### Description

Adds `verifyCliProtocolVersion()` called from `SnykLanguageServer.start()` after the CLI path is validated. If the CLI reports a protocol version that doesn't match `REQUIRED_LS_PROTOCOL_VERSION`, the plugin logs and shows a user-visible error and aborts LS startup. Non-parseable output is treated as non-fatal (logged, startup continues).

Matches the equivalent gate added in VSCode extension PR #733. Part of IDE-1652.

### Checklist

- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

N/A — no UI change.


acke:
Would it be possible to add a smoke test that actually downloads a real Snyk CLI binary and exercises `verifyCliProtocolVersion()` against it end-to-end? The current tests stub the CLI with a temp shell script (and skip on Windows), so they don't catch the case where the real CLI's `language-server --protocolVersion` output format changes or the version-negotiation contract drifts between the plugin and the actual CLI release.

acke (approved):

PR #394, opened 2026-05-18
feat(IDE-1652): introduce LsSettingsRegistry, protocol message objects (1/3)

nick-y-snyk:
## Summary

First of three PRs splitting `feat/IDE-1652`. Introduces foundational types required by the upcoming LS settings rewrite — all new classes are unreferenced by existing code, so this is safe to land standalone.

- **`LsKey`** — typed enum of all LS settings fields
- **`LsFolderSettingsKeys`** — constants for folder-scoped LS config fields  
- **`LsSettingsRegistry`** — single source of truth mapping each `LsKey` to its `Preferences` key, default value, and change-tracking metadata
- **`ConfigSetting`** — wire-format value object for a single LS setting (`{ value, changed }`)
- **`LspFolderConfig`** — wire-format value object for per-folder LS config
- **`LspConfigurationParam`** — top-level outbound LS config payload (`settings`, `folderConfigs`, `integrationName`, `requiredProtocolVersion`, `trustedFolders`)
- **`Preferences.ACTIVATE_SNYK_SECRETS`** — constant needed by `LsSettingsRegistry` (additive, one line)
- **`LsBinaries`** — minor update

Tests cover each value object and serialisation round-trips via Gson.

## PR sequence

| PR | Scope |
|----|-------|
| **This PR (1/3)** | Foundation: registry, protocol POJOs |
| [2/3] | Wiring: inbound `$/snyk.configuration`, outbound `LspConfigurationParam`, `FolderConfigSettings`, change tracking |
| [3/3] | HTML settings page migration onto new store |
| [PR #393](https://github.com/snyk/snyk-eclipse-plugin/pull/393) | All deletions: legacy POJOs + native UI |

## Test plan

- [ ] `./mvnw test -pl plugin,tests` passes
- [ ] `./mvnw pmd:check -pl plugin` — zero violations

___

### **PR Type**
Enhancement

___

### **Description**
- Introduce `LsKey` enum and `LsSettingsRegistry` for centralized setting management.

- Add protocol message objects: `ConfigSetting`, `LspFolderConfig`, `LspConfigurationParam`.

- Establish foundation for Language Server settings rewrite.

- Include unit tests and dependency for new data structures.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A[LsKey Enum] --> B(LsSettingsRegistry);
  B --> C(ConfigSetting);
  B --> D(LspFolderConfig);
  B --> E(LspConfigurationParam);
  F[Preferences.java Add] --> B;
[...]

PR #393, opened 2026-05-15
refactor(IDE-1652): remove native settings UI and legacy FolderConfig infrastructure

nick-y-snyk:
## Summary

Mostly deletions — removes all code made dead by the `LsSettingsRegistry` / `FolderConfigSettings` / `LspConfigurationParam` architecture landed in PRs #394, #395, #396. Also updates 3 files to remove their references to the deleted types (`SnykPreferencePageFactory`, `ProjectPropertyPageFactory`, `SnykExtendedLanguageClient`).

**Breaking behaviour change:** The native Eclipse preference and project property UIs (`PreferencesPage`, `NativeProjectPropertyPage`) are removed. HTML settings are now mandatory — the `SNYK_USE_HTML_SETTINGS` env gate and `isNewConfigDialogEnabled()` flag are dead and also removed.

**Native settings UI** (no callers after HTML-only migration in #396):
- `PreferencesPage`, `TokenFieldEditor`, `NativeProjectPropertyPage`
- `SnykPreferencePageFactoryTest`, `ProjectPropertyPageFactoryTest`

**Legacy LS protocol POJOs** (replaced by `LspConfigurationParam` / `ConfigSetting` / `LspFolderConfig` / `FolderConfigSettings`):
- `Settings`, `FolderConfig`, `FolderConfigsParam`, `FolderConfigs`, `IdeConfigData`
- `FolderConfigTest`

## PR sequence (merge in order)

| PR | Scope |
|----|-------|
| [1/3 — #394](https://github.com/snyk/snyk-eclipse-plugin/pull/394) | Foundation: `LsSettingsRegistry`, protocol POJOs |
| [2/3 — #395](https://github.com/snyk/snyk-eclipse-plugin/pull/395) | Wiring: inbound handler, outbound updater, `FolderConfigSettings` |
| [3/3 — #396](https://github.com/snyk/snyk-eclipse-plugin/pull/396) | HTML settings page migration |
| **This PR** | All deletions (merge last) |

## Test plan

- [ ] `./mvnw test -pl plugin,tests` passes
- [ ] `./mvnw pmd:check -pl plugin` — zero violations

PR #395, opened 2026-05-18
feat(IDE-1652): add inbound $/snyk.configuration handler and outbound updater (2/3)

nick-y-snyk:
## Summary

Second of three PRs splitting `feat/IDE-1652`. Wires up the new LS settings architecture introduced in PR #394. Legacy POJOs (`Settings`, `FolderConfig`, `FolderConfigs`, `FolderConfigsParam`, `IdeConfigData`) are left on disk as dead code — PR #393 deletes them.

- **`FolderConfigSettings`** — per-project store for folder-scoped LS config with explicit change tracking (`storeAndTrackChange`) and batch-flush to `Preferences` on save
- **`LsConfigurationUpdater`** — rewritten to build `LspConfigurationParam` via `LsSettingsRegistry`; replaces legacy `getCurrentSettings()` / `Settings` path
- **`SnykLanguageServer.getInitializationOptions`** — calls `buildConfigurationParam()`, pre-serializes via `Gson.toJsonTree()` so LSP4J embeds JSON correctly for custom POJOs like `ConfigSetting`
- **`SnykExtendedLanguageClient`** — adds `$/snyk.configuration` notification handler (`persistGlobalSettings`) for enterprise config push; outbound path uses `LspConfigurationParam`
- **`LsConstants`** — renames `SNYK_FOLDER_CONFIG` → `SNYK_CONFIGURATION`
- **`Preferences`** — registry-driven defaults, explicit change set, new `storeAndTrackChange` / `getExplicitChanges` / `persistGlobalSettings` APIs; adds `EXPLICIT_CHANGES_KEY`
- **`ExecuteCommandBridge`** — delta for new settings bridge commands
- **`NativeProjectPropertyPage` / `TokenFieldEditor` / `PreferencesPage`** — migrated to `FolderConfigSettings` / `LsSettingsRegistry` so they compile until PR #393 removes them
- **UI callers** (`ReferenceChooserDialog`, `SnykToolView`, `SummaryBrowserHandler`, `BaseHandler`, `SnykWizard`) — migrated to `FolderConfigSettings`

## PR sequence

| PR | Scope |
|----|-------|
| [1/3 — PR #394](https://github.com/snyk/snyk-eclipse-plugin/pull/394) | Foundation: registry, protocol POJOs |
| **This PR (2/3)** | Wiring: inbound handler, outbound updater, folder settings store |
| [3/3] | HTML settings page migration |
| [PR #393](https://github.com/snyk/snyk-eclipse-plugin/pull/393) | All deletions: legacy POJOs + native UI |

## Test plan

- [ ] `./mvnw test -pl plugin,tests` passes
- [ ] `./mvnw pmd:check -pl plugin` — zero violations
- [ ] Smoke: LS starts, `getInitializationOptions` returns `JsonElement` with `settings` / `folderConfigs` / `integrationName` / `requiredProtocolVersion` keys

### DIFF
--- a/plugin/src/main/java/io/snyk/eclipse/plugin/html/ExecuteCommandBridge.java
+++ b/plugin/src/main/java/io/snyk/eclipse/plugin/html/ExecuteCommandBridge.java
@@ -6,6 +6,8 @@ import io.snyk.eclipse.plugin.preferences.AuthConstants;
 import io.snyk.eclipse.plugin.preferences.Preferences;
 import io.snyk.eclipse.plugin.utils.SnykLogger;
 import io.snyk.languageserver.CommandHandler;
+import io.snyk.languageserver.LsKey;
+import io.snyk.languageserver.LsSettingsRegistry;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
@@ -111,9 +113,18 @@ public class ExecuteCommandBridge {
     String endpoint = args.get(1) != null ? String.valueOf(args.get(1)) : "";
     String insecure = String.valueOf(args.get(2));
     Preferences prefs = Preferences.getInstance();
-    prefs.store(Preferences.AUTHENTICATION_METHOD, authMethod);
-    prefs.store(Preferences.ENDPOINT_KEY, endpoint);
-    prefs.store(Preferences.INSECURE_KEY, insecure);
+    storeAndMarkIfChanged(prefs, LsKey.AUTHENTICATION_METHOD, authMethod);
+    storeAndMarkIfChanged(prefs, LsKey.ENDPOINT, endpoint);
+    storeAndMarkIfChanged(prefs, LsKey.INSECURE, insecure);
+    prefs.flushExplicitChanges();
+  }
+
+  private static void storeAndMarkIfChanged(Preferences prefs, LsKey lsKey, String value) {
+    LsSettingsRegistry.Entry entry = LsSettingsRegistry.ENTRIES.get(lsKey);
+    prefs.store(entry.prefKey, value);
+    if (!value.equals(entry.outboundDefault)) {
+      prefs.markExplicitlyChangedNoFlush(entry.prefKey);
+    }
   }
 
   private static void registerBridgeFunction(Browser browser) {
--- a/plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
+++ b/plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
@@ -129,7 +129,7 @@ public class Preferences {
 		// Defaults for LS-backed keys driven from registry — single source of truth.
 		// Skips: fixed entries (no prefKey), AUTH_TOKEN_KEY (encrypted, not in insecureStore), CLI_PATH (computed below).
 		for (LsSettingsRegistry.Entry entry : LsSettingsRegistry.ENTRIES.values()) {
-			if (entry.prefKey == null || AUTH_TOKEN_KEY.equals(entry.prefKey) || CLI_PATH.equals(entry.prefKey)) {
+			if (entry.prefKey == null || entry.encrypted || CLI_PATH.equals(entry.prefKey)) {
 				continue;
 			}
 			insecureStore.setDefault(entry.prefKey, entry.outboundDefault);
@@ -142,6 +142,7 @@ public class Preferences {
 		insecureStore.setDefault(ANALYTICS_PLUGIN_INSTALLED_SENT, FALSE);
 		insecureStore.setDefault(DEVICE_ID, UUID.randomUUID().toString());
 		insecureStore.setDefault(USE_LS_HTML_CONFIG_DIALOG, TRUE);
+		// TODO: move to LsSettingsRegistry once Entry supports Supplier<String> defaults (CLI_PATH is runtime-computed)
 		insecureStore.setDefault(CLI_PATH, getDefaultCliPath());
 
 		String savedExplicitChanges = insecure.get(EXPLICIT_CHANGES_KEY, "");
@@ -197,6 +198,21 @@ public class Preferences {
 		if (getBooleanPref(USE_TOKEN_AUTH)) {
 			store(AUTHENTICATION_METHOD, AuthConstants.AUTH_API_TOKEN);
 			store(USE_TOKEN_AUTH, FALSE);
+		}
+
+		// One-time upgrade migration: seed explicitChanges for any pref already
+		// customized on disk, so the first outbound sync sends changed=true for them.
+		if (explicitChanges.isEmpty()) {
+			for (LsSettingsRegistry.Entry entry : LsSettingsRegistry.ENTRIES.values()) {
+				if (entry.prefKey == null || entry.encrypted) {
+					continue;
+				}
+				String stored = insecurePreferences.get(entry.prefKey, null);
+				if (stored != null && !stored.equals(entry.outboundDefault)) {
+					explicitChanges.add(entry.prefKey);
+				}
+			}
+			persistExplicitChanges();
 		}
 	}
 
--- a/plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
+++ b/plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
@@ -9,6 +9,7 @@ import java.io.File;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
@@ -28,6 +29,7 @@ import io.snyk.eclipse.plugin.Activator;
 import io.snyk.eclipse.plugin.EnvironmentConstants;
 import io.snyk.eclipse.plugin.utils.SnykLogger;
 import io.snyk.languageserver.LsRuntimeEnvironment;
+import io.snyk.languageserver.LsSettingsRegistry;
 
 public class Preferences {
 	private static final String FALSE = "false";
@@ -85,6 +87,7 @@ public class Preferences {
 	public static final String DEVICE_ID = "deviceId";
 	public static final String RELEASE_CHANNEL = "releaseChannel";
 	public static final String USE_LS_HTML_CONFIG_DIALOG = "useLsHtmlConfigDialog";
+	public static final String EXPLICIT_CHANGES_KEY = "explicitChanges";
 
 	private static final Set<String> encryptedPreferenceKeys = Set.of(AUTH_TOKEN_KEY);
 	private final IEclipsePreferences insecurePreferences;
@@ -93,6 +96,7 @@ public class Preferences {
 	private IPreferenceStore secureStore;
 	private boolean secureStorageReady;
 	private Map<String, String> prefSaveMap = new ConcurrentHashMap<>();
+	private final Set<String> explicitChanges = ConcurrentHashMap.newKeySet();
 	private static Function<String, String> envProvider = System::getenv;
 
 	public static synchronized Preferences getInstance() {
@@ -122,32 +126,33 @@ public class Preferences {
 			waitForSecureStorage();
 		});
 
-		insecureStore.setDefault(ACTIVATE_SNYK_CODE_SECURITY, FALSE);
-		insecureStore.setDefault(ACTIVATE_SNYK_OPEN_SOURCE, TRUE);
-		insecureStore.setDefault(ACTIVATE_SNYK_IAC, TRUE);
-		insecureStore.setDefault(FILTER_SHOW_CRITICAL, TRUE);
-		insecureStore.setDefault(FILTER_SHOW_HIGH, TRUE);
-		insecureStore.setDefault(FILTER_SHOW_MEDIUM, TRUE);
-		insecureStore.setDefault(FILTER_SHOW_LOW, TRUE);
-		insecureStore.setDefault(ENABLE_DELTA, FALSE);
-		insecureStore.setDefault(RISK_SCORE_THRESHOLD, "0");
-		insecureStore.setDefault(FILTER_IGNORES_SHOW_OPEN_ISSUES, TRUE);
-		insecureStore.setDefault(FILTER_IGNORES_SHOW_IGNORED_ISSUES, FALSE);
+		// Defaults for LS-backed keys driven from registry — single source of truth.
+		// Skips: fixed entries (no prefKey), AUTH_TOKEN_KEY (encrypted, not in insecureStore), CLI_PATH (computed below).
+		for (LsSettingsRegistry.Entry entry : LsSettingsRegistry.ENTRIES.values()) {
+			if (entry.prefKey == null || AUTH_TOKEN_KEY.equals(entry.prefKey) || CLI_PATH.equals(entry.prefKey)) {
+				continue;
+			}
+			insecureStore.setDefault(entry.prefKey, entry.outboundDefault);
+		}
+		// Non-LS keys with no registry entry.
 		insecureStore.setDefault(FILTER_SHOW_ONLY_FIXABLE, FALSE);
-		insecureStore.setDefault(MANAGE_BINARIES_AUTOMATICALLY, TRUE);
 		insecureStore.setDefault(LSP_VERSION, "1");
 		insecureStore.setDefault(IS_GLOBAL_IGNORES_FEATURE_ENABLED, FALSE);
-		insecureStore.setDefault(CLI_BASE_URL, "https://downloads.snyk.io");
-		insecureStore.setDefault(SCANNING_MODE_AUTOMATIC, TRUE);
 		insecureStore.setDefault(USE_TOKEN_AUTH, FALSE);
-		insecureStore.setDefault(AUTHENTICATION_METHOD, AuthConstants.AUTH_OAUTH2);
 		insecureStore.setDefault(ANALYTICS_PLUGIN_INSTALLED_SENT, FALSE);
 		insecureStore.setDefault(DEVICE_ID, UUID.randomUUID().toString());
-		insecureStore.setDefault(RELEASE_CHANNEL, "stable");
 		insecureStore.setDefault(USE_LS_HTML_CONFIG_DIALOG, TRUE);
 		insecureStore.setDefault(CLI_PATH, getDefaultCliPath());
-		insecureStore.setDefault(ENDPOINT_KEY, DEFAULT_ENDPOINT);
-		insecureStore.setDefault(ORGANIZATION_KEY, "");
+
+		String savedExplicitChanges = insecure.get(EXPLICIT_CHANGES_KEY, "");
+		if (savedExplicitChanges != null && !savedExplicitChanges.isEmpty()) {
+			for (String key : savedExplicitChanges.split(",")) {
+				String trimmed = key.trim();
+				if (!trimmed.isEmpty()) {
+					explicitChanges.add(trimmed);
+				}
+			}
+		}
 
 		var endpoint = getEnvironmentVariable(ENV_SNYK_API, "");
 		if (endpoint != null && !endpoint.isBlank()) {
@@ -374,5 +379,66 @@ public class Preferences {
 			return false;
 		}
 		return true;
+	}
+
+	public final void storeAndTrackChange(String key, String value) {
+		if (key == null || value == null) {
+			return;
+		}
+		String oldValue = getPref(key);
+		store(key, value);
+		if (!Objects.equals(oldValue, value)) {
+			markExplicitlyChanged(key);
+		}
+	}
+
+	public void markExplicitlyChanged(String key) {
+		if (key == null) {
+			return;
+		}
+		explicitChanges.add(key);
+		persistExplicitChanges();
+	}
+
+	public void markExplicitlyChangedNoFlush(String key) {
+		if (key == null) {
+			return;
+		}
+		explicitChanges.add(key);
+	}
+
+	public void clearExplicitlyChangedNoFlush(String key) {
+		if (key == null) {
+			return;
+		}
+		explicitChanges.remove(key);
+	}
+
+	public void flushExplicitChanges() {
+		persistExplicitChanges();
+	}
+
+	public boolean isExplicitlyChanged(String key) {
+		if (key == null) {
+			return false;
+		}
+		return explicitChanges.contains(key);
+	}
+
+	public void clearExplicitlyChanged(String key) {
+		if (key == null) {
+			return;
+		}
+		explicitChanges.remove(key);
+		persistExplicitChanges();
+	}
+
+	private void persistExplicitChanges() {
+		insecurePreferences.put(EXPLICIT_CHANGES_KEY, String.join(",", explicitChanges));
+		try {
+			insecurePreferences.flush();
+		} catch (org.osgi.service.prefs.BackingStoreException e) {
+			SnykLogger.logError(e);
+		}
 	}
 }
--- a/plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java
+++ b/plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java
@@ -32,7 +32,7 @@ public class FolderConfigSettings {
 		instance = i;
 	}
 
-	public void addFolderConfig(LspFolderConfig folderConfig) {
+	public synchronized void addFolderConfig(LspFolderConfig folderConfig) {
 		if (folderConfig.getFolderPath() == null) {
 			return;
 		}
@@ -90,7 +90,7 @@ public class FolderConfigSettings {
 		return configs.containsKey(key);
 	}
 
-	public void updateFolderConfig(String folderPath, LspFolderConfig config) {
+	public synchronized void updateFolderConfig(String folderPath, LspFolderConfig config) {
 		if (folderPath == null) {
 			return;
 		}
--- a/plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java
+++ b/plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java
@@ -1,0 +1,196 @@
+package io.snyk.eclipse.plugin.properties;
+
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+
+import io.snyk.eclipse.plugin.utils.SnykLogger;
+import io.snyk.languageserver.LsFolderSettingsKeys;
+import io.snyk.languageserver.protocolextension.messageObjects.ConfigSetting;
+import io.snyk.languageserver.protocolextension.messageObjects.LspFolderConfig;
+
+public class FolderConfigSettings {
+
+	private static FolderConfigSettings instance;
+
+	private final Map<String, LspFolderConfig> configs = new ConcurrentHashMap<>();
+
+	public static synchronized FolderConfigSettings getInstance() {
+		if (instance == null) {
+			instance = new FolderConfigSettings();
+		}
+		return instance;
+	}
+
+	public static synchronized void setInstance(FolderConfigSettings i) {
+		instance = i;
+	}
+
+	public void addFolderConfig(LspFolderConfig folderConfig) {
+		if (folderConfig.getFolderPath() == null) {
+			return;
+		}
+		String key = normalizePath(folderConfig.getFolderPath());
+		configs.put(key, folderConfig);
+	}
+
+	/**
+	 * Replaces the full folder config state with the incoming snapshot from the LS.
+	 * LS protocol v25 guarantees {@code $/snyk.configuration} carries all folders — incoming list
+	 * is authoritative; folders absent from it are removed.
+	 */
+	public synchronized void addAll(List<LspFolderConfig> folderConfigs) {
+		// Collect paths present in the incoming list; remove folders no longer reported by LS.
+		Set<String> incomingKeys = new HashSet<>();
+		for (LspFolderConfig incoming : folderConfigs) {
+			if (incoming.getFolderPath() == null) {
+				continue;
+			}
+			String key = normalizePath(incoming.getFolderPath());
+			incomingKeys.add(key);
+			configs.put(key, incoming);
+		}
+		configs.keySet().retainAll(incomingKeys);
+	}
+
+	public LspFolderConfig getFolderConfig(String folderPath) {
+		if (folderPath == null) {
+			return createEmptyConfig(null);
+		}
+		String key = normalizePath(folderPath);
+		LspFolderConfig config = configs.get(key);
+		if (config == null) {
+			return createEmptyConfig(folderPath);
+		}
+		return config;
+	}
+
+	private LspFolderConfig createEmptyConfig(String folderPath) {
+		LspFolderConfig config = new LspFolderConfig();
+		config.setFolderPath(folderPath);
+		config.setSettings(new java.util.HashMap<>());
+		return config;
+	}
+
+	public synchronized List<LspFolderConfig> getAll() {
+		return new ArrayList<>(configs.values());
+	}
+
+	public boolean isConfigured(String folderPath) {
+		if (folderPath == null) {
+			return false;
+		}
+		String key = normalizePath(folderPath);
+		return configs.containsKey(key);
+	}
+
+	public void updateFolderConfig(String folderPath, LspFolderConfig config) {
+		if (folderPath == null) {
+			return;
+		}
+		String key = normalizePath(folderPath);
+		configs.put(key, config);
+	}
+
+	public synchronized void computeFolderConfig(String folderPath, Function<LspFolderConfig, LspFolderConfig> updater) {
+		if (folderPath == null) {
+			return;
+		}
+		String key = normalizePath(folderPath);
+		LspFolderConfig existing = configs.get(key);
+		if (existing == null) {
+			existing = createEmptyConfig(folderPath);
+		}
+		configs.put(key, updater.apply(existing));
+	}
+
+	public String getBaseBranch(String path) {
+		return getStringSettingValue(path, LsFolderSettingsKeys.BASE_BRANCH);
+	}
+
+	public String getPreferredOrg(String path) {
+		return getStringSettingValue(path, LsFolderSettingsKeys.PREFERRED_ORG);
+	}
+
+	public List<String> getLocalBranches(String path) {
+		ConfigSetting setting = getSetting(path, LsFolderSettingsKeys.LOCAL_BRANCHES);
+		if (setting == null || setting.getValue() == null) {
+			return Collections.emptyList();
+		}
+		Object value = setting.getValue();
+		if (value instanceof List<?> rawList) {
+			List<String> result = new ArrayList<>();
+			for (Object item : rawList) {
+				result.add(String.valueOf(item));
+			}
+			return result;
+		}
+		return Collections.emptyList();
+	}
+
+	public List<String> getAdditionalParameters(String path) {
+		ConfigSetting setting = getSetting(path, LsFolderSettingsKeys.ADDITIONAL_PARAMETERS);
+		if (setting == null || setting.getValue() == null) {
+			return Collections.emptyList();
+		}
+		Object value = setting.getValue();
+		if (value instanceof List<?> rawList) {
+			List<String> result = new ArrayList<>();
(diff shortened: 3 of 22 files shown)


tests/src/test/java/io/snyk/languageserver/SnykLanguageServerTest.java
```diff
@@ -87,4 +121,5 @@ void getCliPathOrThrow_returnsPathWhenBinaryExists() throws IOException {
 
     assertEquals(existingBinary.getAbsolutePath(), result);
   }
+
```
acke: Remove extra linebreak

acke (changes requested):
## Orchestrated review (review-orchestrator)

**Verdict: REQUEST CHANGES**

Wiring aligns with protocol v25 and test coverage is strong on happy paths. Merge blockers and should-fix items are in inline threads (reformatted) and summarized below.

---

### 🔴 Critical

**Preferences.java:147-201** — No upgrade migration seeds `explicitChanges` for existing customized prefs; first LS sync sends `changed=false`.

**Fix:**
One-time migration in constructor/`migratePreferences()` marking non-default stored prefs explicit. Test: upgrade fixture with custom endpoint → `buildConfigurationParam` sends `changed=true`.

— AI review comment

**SnykExtendedLanguageClient.java:438-453** — Inbound `$/snyk.configuration` persists `token`; VS Code skips it.

**Fix:**
Skip `LsKey.TOKEN` in `persistGlobalSettings`, matching VS Code.

— AI review comment

**LsSettingsRegistry.java:124-127** — `scan_automatic` inbound mishandles boolean `true`.

**Fix:**
Accept `Boolean.TRUE`, `"true"`, and `"automatic"` in inbound deserializer; add boolean inbound test.

— AI review comment

**SnykExtendedLanguageClient.java:319-320** — `hasAuthenticated` NPE on null token.

**Fix:**
Use `Objects.equals(newToken, oldToken)`.

— AI review comment

**ExecuteCommandBridge.java:169-189** — DOM XSS via `browser.evaluate` + raw JSON.

**Fix:**
Use `BrowserFunction` return value or safe encoding; never splice raw JSON.

— AI review comment

---

### 🟠 High / 🟡 Medium

See inline threads (each prefixed with severity, **Fix:** block, then `— AI review comment`).

---

*Manifest: `.claude/reviews/orchestrator.Ry0HZf/TOOLS_USED.md`*

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -409,21 +409,47 @@ public void snykScan(SnykScanParam param) {
...
+		var prefs = Preferences.getInstance();
+		for (var entry : settings.entrySet()) {
+			try {
+				var registryEntry = io.snyk.languageserver.LsSettingsRegistry.BY_LS_KEY.get(entry.getKey());
+				if (registryEntry == null || registryEntry.prefKey == null) {
+					continue;
+				}
+				var setting = entry.getValue();
+				if (setting.getValue() != null) {
+					prefs.store(registryEntry.prefKey, registryEntry.inboundDeserializer.apply(setting.getValue()));
```
acke, line 459: 🔴 **Critical**

Inbound `$/snyk.configuration` persists any registry key with a `prefKey`, including `token`. VS Code skips `LS_GLOBAL_KEY.token` (no `vscodeKey` in `lsKeyToVscodeKeyMap.ts:131-133, 277-283`). A compromised LS can overwrite `AUTH_TOKEN_KEY`.

**Fix:**
Skip `LsKey.TOKEN` (and other LS-only keys) in `persistGlobalSettings`, matching VS Code.

— AI review comment

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
```diff
@@ -122,32 +126,33 @@ public static synchronized Preferences getTestInstance(IEclipsePreferences insec
...
 		insecureStore.setDefault(CLI_PATH, getDefaultCliPath());
-		insecureStore.setDefault(ENDPOINT_KEY, DEFAULT_ENDPOINT);
-		insecureStore.setDefault(ORGANIZATION_KEY, "");
+
+		String savedExplicitChanges = insecure.get(EXPLICIT_CHANGES_KEY, "");
+		if (savedExplicitChanges != null && !savedExplicitChanges.isEmpty()) {
+			for (String key : savedExplicitChanges.split(",")) {
+				String trimmed = key.trim();
+				if (!trimmed.isEmpty()) {
+					explicitChanges.add(trimmed);
```
acke, line 153: 🔴 **Critical**

No upgrade migration seeds `explicitChanges` for prefs already customized on disk. After upgrade, the first LS sync sends `changed=false` for endpoint, org, filters, delta, products, etc., so LS may ignore existing IDE settings.

**Fix:**
Add a one-time migration in the constructor or `migratePreferences()` that marks non-default stored prefs as explicit. Test first with an upgrade fixture that has a custom endpoint and assert `buildConfigurationParam` sends `changed=true`.

— AI review comment

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -409,21 +409,47 @@ public void snykScan(SnykScanParam param) {
...
+
+			if (param.getSettings() != null) {
+				persistGlobalSettings(param.getSettings());
+			}
+
+			if (param.getFolderConfigs() != null) {
+				var folderConfigSettings = FolderConfigSettings.getInstance();
+				folderConfigSettings.addAll(param.getFolderConfigs());
+				if (Preferences.getInstance().getBooleanPref(Preferences.SCANNING_MODE_AUTOMATIC)) {
+					triggerScan(null);
```
acke, line 435: 🟠 **High**

Folder inbound notifications call `triggerScan(null)` (workspace scan) instead of per-folder `triggerScan(path)`. Any folder update can trigger a full-workspace scan.

**Fix:**
Restore per-folder scan when automatic mode is on, or confirm LS treats a workspace scan as equivalent.

— AI review comment

plugin/src/main/java/io/snyk/eclipse/plugin/html/ExecuteCommandBridge.java
```diff
@@ -111,9 +113,18 @@ static void saveLoginArgs(List<Object> args) {
...
+    storeAndMarkIfChanged(prefs, LsKey.AUTHENTICATION_METHOD, authMethod);
+    storeAndMarkIfChanged(prefs, LsKey.ENDPOINT, endpoint);
+    storeAndMarkIfChanged(prefs, LsKey.INSECURE, insecure);
+    prefs.flushExplicitChanges();
+  }
+
+  private static void storeAndMarkIfChanged(Preferences prefs, LsKey lsKey, String value) {
+    LsSettingsRegistry.Entry entry = LsSettingsRegistry.ENTRIES.get(lsKey);
+    prefs.store(entry.prefKey, value);
+    if (!value.equals(entry.outboundDefault)) {
```
acke, line 125: 🟠 **High**

`storeAndMarkIfChanged` only marks a pref explicit when value ≠ registry default. Reverting to default does not clear `explicitChanges`, and defaults saved from HTML/wizard never mark explicit — outbound sync sends `changed=false`.

**Fix:**
Align with `Preferences.storeAndTrackChange` (compare prior value; clear explicit on revert to default).

— AI review comment

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
```diff
@@ -112,13 +114,16 @@ public Object getInitializationOptions(URI rootUri) {
...
+			// Pre-serialize to JsonElement so LSP4J embeds the JSON directly
+			// instead of re-serializing through its own Gson instance, which
+			// can drop fields from custom POJOs like ConfigSetting.
+			return new Gson().toJsonTree(param);
+		} catch (Exception e) { // NOPMD - broad catch prevents LS handshake abort on unexpected errors
 			// Handle initialization errors gracefully - log and return null to allow LS to start
 			SnykLogger.logError(e);
 		}
-		return currentSettings;
+		return null;
```
acke, line 127: 🟠 **High**

`getInitializationOptions` returns `null` on any serialization failure. LS may start with no IDE settings.

**Fix:**
Return a minimal safe `JsonElement` or last-known-good payload. Add a failure-path test.

— AI review comment

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
acke, line 459: 🟠 **High**

Inbound settings with `value: null` are skipped. LS cannot clear or reset a stored pref from an inbound push.

**Fix:**
Define clear semantics (remove key or write default) when LS sends `null` with `changed: true`.

— AI review comment

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
```diff
@@ -24,74 +20,60 @@ public void configurationChanged() {
...
-
-		return new Settings(activateSnykOpenSource, activateSnykCodeSecurity, activateSnykIac,
-				insecure, endpoint, additionalParams, additionalEnv, path, issueViewOptions, sendErrorReports,
-				enableTelemetry, organization, manageBinariesAutomatically, cliPath, cliBaseDownloadURL, token,
-				integrationName, integrationVersion, automaticAuthentication, trustedFolders, enableTrustedFolderFeature,
-				scanningMode, enableDeltaFindings, riskScoreThreshold, authenticationMethod, folderConfigs, filterSeverity);
+		String[] trustedFolders = (String[]) LsSettingsRegistry.ENTRIES.get(LsKey.TRUSTED_FOLDERS)
+				.outboundSerializer.apply(preferences.getPref(Preferences.TRUSTED_FOLDERS, ""));
+
+		var folderConfigs = FolderConfigSettings.getInstance().getAll();
```
acke, line 60: 🟠 **High**

`HTMLSettingsPreferencePage.java:259-287` still reads/writes legacy `FolderConfigs`, while outbound uses `FolderConfigSettings.getAll()` here. HTML-saved folder config never reaches LS after this PR removes the `folderConfig` handler.

**Fix:**
Migrate the HTML page to `FolderConfigSettings` in PR 3, or document as a known gap with env-gated HTML only.

— AI review comment

plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java
```diff
@@ -0,0 +1,196 @@
...
+			instance = new FolderConfigSettings();
+		}
+		return instance;
+	}
+
+	public static synchronized void setInstance(FolderConfigSettings i) {
+		instance = i;
+	}
+
+	public void addFolderConfig(LspFolderConfig folderConfig) {
```
acke: 🟡 **Medium**

`addFolderConfig` is unsynchronized while `addAll` uses `synchronized` + `retainAll`. Concurrent LS snapshot and UI `computeFolderConfig` can race.

**Fix:**
Use a single locking model for all mutators.

— AI review comment

tests/src/test/java/io/snyk/eclipse/plugin/properties/preferences/PreferencesTest.java
```diff
@@ -209,4 +209,173 @@ public void testIsNewConfigDialogEnabled_falseWhenEnvVarSetToFalse() {
 		assertFalse(Preferences.isNewConfigDialogEnabled());
 	}
 
+	@Test
+	public void testMarkExplicitlyChanged_addsKeyToSet() {
```
acke, line 196: 🟡 **Medium**

`PreferencesTest.java:51` — `test_DefaultPreferences` still asserts hardcoded defaults while the constructor now seeds from `LsSettingsRegistry`. The test may pass coincidentally but will not catch registry drift.

**Fix:**
Update assertions to registry defaults, or remove obsolete expectations.

— AI review comment

tests/src/test/java/io/snyk/languageserver/LsConfigurationUpdaterTest.java
```diff
@@ -2,91 +2,413 @@
...
+		FolderConfigSettings.setInstance(folderConfigSettings);
+
+		var param = new LsConfigurationUpdater().buildConfigurationParam();
+		List<LspFolderConfig> folderConfigs = param.getFolderConfigs();
+
+		assertEquals(1, folderConfigs.size());
+	}
+
+	@Test
+	void testRoundTrip_userChangeProducesCorrectPayload() {
```
acke, line 173: 🟡 **Medium**

Removed integration-style `testGetSettings`. Replacement tests mostly assert `changed` flags, not outbound values/types (risk score, severities).

**Fix:**
Restore value assertions for high-risk keys, or add more round-trip tests like `testRoundTrip_userChangeProducesCorrectPayload`.

— AI review comment

acke:
## Review format update — orchestrated review

Inline comments have been reformatted: **severity label**, issue description, **Fix:** on its own block, then `— AI review comment` on a separate line.

### 🔴 Critical (additional — not inline; file/lines outside this PR diff)

**LsSettingsRegistry.java:124-127** — `scan_automatic` inbound deserializer only treats `"automatic"` as true; LS boolean `true` becomes pref `"false"`.

**Fix:**
Accept `Boolean.TRUE`, `"true"`, and `"automatic"` in the inbound deserializer. Add a test with boolean inbound value.

— AI review comment

---

**SnykExtendedLanguageClient.java:319-320** — `hasAuthenticated` uses `!newToken.equals(oldToken)` with no null guard; logout/null token can NPE.

**Fix:**
Use `Objects.equals(newToken, oldToken)` before UI/store updates.

— AI review comment

---

**ExecuteCommandBridge.java:169-189** — Raw `resultJson` embedded in `browser.evaluate` (DOM XSS risk).

**Fix:**
Pass results via `BrowserFunction` return value or safe encoding; never splice raw JSON into evaluated script.

— AI review comment

---

**Verdict: REQUEST CHANGES** — Critical items are labeled `🔴 **Critical**` in updated inline threads (Preferences `explicitChanges` migration, inbound `token` persistence).

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Fixed via allowlist pattern (matching IntelliJ): `persistGlobalSettings` now only persists keys in `INBOUND_ALLOWED_GLOBAL_KEYS` — TOKEN is absent, so LS can never overwrite auth credentials.

*— Claude Code*

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Won't fix — there's no per-folder inbound notification in this protocol version. `triggerScan(null)` fires `COMMAND_WORKSPACE_SCAN`, which the LS routes to `ScanWorkspace()` → iterates all trusted folders and scans each. Verified in `workspace.go:216-221`.

*— Claude Code*

plugin/src/main/java/io/snyk/eclipse/plugin/html/ExecuteCommandBridge.java
nick-y-snyk: Won't fix now — reset-to-default functionality isn't implemented yet. When it is, `storeAndMarkIfChanged` will need to clear `explicitChanges` on revert. Tracked as a known gap.

*— Claude Code*

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
nick-y-snyk: Won't fix. Returning `null` is valid LSP — it signals no `initializationOptions`, and the LS starts with its own defaults. Eclipse immediately follows with `didChangeConfiguration` which delivers the real config. IntelliJ and VS Code have no null fallback either — they rely on safe defaults in the settings builders. If `buildConfigurationParam()` throws, the same failure would hit the subsequent `didChangeConfiguration` call anyway.

*— Claude Code*

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Won't fix now — reset functionality not yet implemented. When it is, null+changed:true will need defined semantics (write default or remove key).

*— Claude Code*

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
nick-y-snyk: Addressed in PR #396 (3/3) which migrates `HTMLSettingsPreferencePage` to `FolderConfigSettings`.

*— Claude Code*

plugin/src/main/java/io/snyk/eclipse/plugin/properties/FolderConfigSettings.java
nick-y-snyk: Fixed: `addFolderConfig` and `updateFolderConfig` are now `synchronized`, consistent with `addAll`, `getAll`, and `computeFolderConfig`.

*— Claude Code*

tests/src/test/java/io/snyk/languageserver/SnykLanguageServerTest.java
nick-y-snyk: Fixed: removed extra blank line before closing brace.

*— Claude Code*

tests/src/test/java/io/snyk/eclipse/plugin/properties/preferences/PreferencesTest.java
nick-y-snyk: Fixed: `test_DefaultPreferences` now iterates `LsSettingsRegistry.ENTRIES` and asserts each pref's stored value matches its `outboundDefault`. Registry drift will now fail the test.

*— Claude Code*

tests/src/test/java/io/snyk/languageserver/LsConfigurationUpdaterTest.java
nick-y-snyk: Added `testBuildConfigurationParam_highRiskKeySerializedValues` covering: `risk_score_threshold` serializes `"200"` → `Integer(200)` and `"0"` → `null`; `scan_automatic` serializes `"true"` → `Boolean.TRUE` and `"false"` → `Boolean.FALSE`. Severity value serialization was already covered at line 336-339.

*— Claude Code*

acke (changes requested):
## Re-review (review-orchestrator)

**Verdict: REQUEST CHANGES**

### Verification of prior comments

| # | Severity | File:Line | Status |
|---|----------|-----------|--------|
| 1 | Nit | `SnykLanguageServerTest.java:124` blank line | ✅ Fixed (`5a0cf25`) |
| 2 | **Critical** | `SnykExtendedLanguageClient.java:467` token persistence | ✅ Fixed (`ff1ac95`) — but see new pass C1 |
| 3 | **Critical** | `Preferences.java:153` upgrade migration | ✅ Fixed (`29327d7`) — no GH reply; please mark resolved |
| 4 | High | `SnykExtendedLanguageClient.java:443` per-folder scan | ⏸️ Declined (protocol limit) |
| 5 | High | `ExecuteCommandBridge.java:125` storeAndMarkIfChanged revert | ⏸️ Deferred — see new pass M3 (add TODO marker) |
| 6 | High | `SnykLanguageServer.java:127` null InitializationOptions | ⏸️ Declined (LSP-valid) |
| 7 | High | `SnykExtendedLanguageClient.java:467` inbound null skipped | ⏸️ Deferred (reset semantics) |
| 8 | High | `LsConfigurationUpdater.java:60` HTML legacy folder config | ⏸️ Deferred to #396 (verified, OK) |
| 9 | Medium | `FolderConfigSettings.java:35` unsynchronized addFolderConfig | ✅ Fixed (`48d2cd4`) |
| 10 | Medium | `PreferencesTest.java:196` stale defaults | ✅ Fixed (`2ff4e34`) |
| 11 | Medium | `LsConfigurationUpdaterTest.java:173` missing value assertions | ✅ Fixed (`2f25652`) |

### New pass findings (7)

- 🔴 **Critical** `SnykExtendedLanguageClient.java:115-123` — `INBOUND_ALLOWED_GLOBAL_KEYS` allows only 7 keys, but the test suite asserts persistence of ~10 more (`scan_automatic`, `snyk_code_enabled`, `snyk_secrets_enabled`, severities, etc.). Either allowlist is over-restrictive (functional regression vs. pre-patch behavior) or the tests pass coincidentally and provide false coverage. **Decide.**
- 🟠 **High** `Preferences.java:203-216` — upgrade migration re-runs if `explicitChanges` ever becomes empty. Use a separate one-shot flag.
- 🟠 **High** `SnykExtendedLanguageClient.java:454-473` — `persistGlobalSettings` runs on lsp4j dispatch thread; wrap in `CompletableFuture.runAsync` like `snykScan` does.
- 🟡 **Medium** `Preferences.java:453` — `explicitChanges` serialized as comma-joined; brittle.
- 🟡 **Medium** `LsConfigurationUpdater.java:47` — `Entry.alwaysChanged` for TOKEN sends `token=""` with `changed=true` when unauthenticated; verify LS doesn't treat as logout.
- 🟡 **Medium** `ExecuteCommandBridge.java:125` — please leave a `TODO(IDE-XXXX)` marker for the deferred reset-to-default semantics.
- 🔵 **Low** `SnykExtendedLanguageClient.java:449` — silent catch on `snykConfiguration`; consider an info log at handler entry.

See inline comments for full block-formatted findings.

Blocking item: the allowlist decision. Once that's resolved, the rest are polish.

---
Manifest: `.claude/reviews/395/TOOLS_USED.md`
— AI review comment

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -91,20 +94,33 @@
...
+import io.snyk.languageserver.LsKey;
+import io.snyk.languageserver.LsSettingsRegistry;
 import io.snyk.languageserver.protocolextension.messageObjects.scanResults.Issue;
 
 @SuppressWarnings({"restriction", "PMD.AvoidCatchingGenericException"})
 public class SnykExtendedLanguageClient extends LanguageClientImpl {
 	private static final String MARKER_TYPE = "io.snyk.languageserver.marker";
+
+	// Keys the LS is allowed to push back to the IDE. TOKEN is write-only (IDE → LS only).
+	private static final Set<LsKey> INBOUND_ALLOWED_GLOBAL_KEYS = EnumSet.of(
```
acke: 🔴 **Critical**

**[SnykExtendedLanguageClient.java:115-123]** `INBOUND_ALLOWED_GLOBAL_KEYS` allows only 7 keys: `ENDPOINT, ORGANIZATION, INSECURE, MANAGE_BINARIES_AUTOMATICALLY, CLI_PATH, CLI_BASE_DOWNLOAD_URL, CLI_RELEASE_CHANNEL`. Every other LS-pushed key is silently dropped. But `SnykExtendedLanguageClientTest.java:700-857` asserts that `snykConfiguration` persists `scan_automatic`, `snyk_code_enabled`, `snyk_secrets_enabled`, `snyk_oss_enabled`, `send_error_reports`, `automatic_download`, `proxy_insecure`, `authentication_method`, and `scan_net_new`. Most of those are NOT in the allowlist — `persistGlobalSettings` skips them.

Two things can't both be true:
- **Either the allowlist is missing keys** → functional regression: LS can never sync IDE state for product toggles, scan mode, severity filters, even when the IDE explicitly requested an inbound push.
- **Or the tests pass coincidentally** (the asserted value happens to match the registry-driven default) → tests are dead weight and don't actually exercise the inbound persistence path.

Either way, the security fix overshot. IntelliJ blocks only `token`, not the entire product/scan/auth surface.

**Fix:**
Decide which keys the LS is allowed to push. Safest starting position is "everything in BY_LS_KEY except the encrypted/auth set", expressed as a denylist:
```java
private static final Set<LsKey> INBOUND_DENIED_GLOBAL_KEYS = EnumSet.of(
    LsKey.TOKEN,
    LsKey.AUTHENTICATION_METHOD,   // if IDE owns this
    LsKey.AUTOMATIC_AUTHENTICATION
);
// then: if (INBOUND_DENIED_GLOBAL_KEYS.contains(registryEntry.lsKey)) continue;
```
Or extend the current allowlist to include every key the tests assert against, and add a unit test that fails when a registry key exists with no corresponding allow/deny decision.

Also add a negative test asserting inbound `token` is dropped — currently no test confirms the security fix.

— AI review comment

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
```diff
@@ -193,6 +199,21 @@ private void migratePreferences() {
 			store(AUTHENTICATION_METHOD, AuthConstants.AUTH_API_TOKEN);
 			store(USE_TOKEN_AUTH, FALSE);
 		}
+
+		// One-time upgrade migration: seed explicitChanges for any pref already
+		// customized on disk, so the first outbound sync sends changed=true for them.
+		if (explicitChanges.isEmpty()) {
```
acke, line 205: 🟠 **High**

**[Preferences.java:203-216]** The upgrade-migration block is gated on `if (explicitChanges.isEmpty())`. If the user (or a future "reset all to defaults" feature) ever clears `explicitChanges` and restarts, this block re-runs and re-marks every non-default pref as explicit — silently overriding the user's intent.

**Fix:**
Use a separate one-shot flag:
```java
public static final String EXPLICIT_CHANGES_MIGRATED = "explicitChangesMigrated";
...
if (\!insecurePreferences.getBoolean(EXPLICIT_CHANGES_MIGRATED, false)) {
    for (LsSettingsRegistry.Entry entry : LsSettingsRegistry.ENTRIES.values()) { ... }
    insecurePreferences.putBoolean(EXPLICIT_CHANGES_MIGRATED, true);
    persistExplicitChanges();
}
```
Add a test: first call seeds; second call (after clearing `explicitChanges`) does not re-seed.

— AI review comment

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -409,21 +425,50 @@ public void snykScan(SnykScanParam param) {
...
+				}
+			}
+		} catch (Exception e) {
+			SnykLogger.logError(e);
 		}
-		if (this.toolView != null) {
-			this.toolView.refreshDeltaReference();
+	}
+
+	private void persistGlobalSettings(java.util.Map<String, ConfigSetting> settings) {
```
acke, line 446: 🟠 **High**

**[SnykExtendedLanguageClient.java:454-473]** `persistGlobalSettings` is invoked synchronously from `snykConfiguration` (the `$/snyk.configuration` `@JsonNotification` handler) — i.e. on the lsp4j dispatch thread. Each `prefs.store(...)` writes to the Eclipse `IEclipsePreferences` backing store, which can block on disk I/O. If a flush stalls, LS RPC pipeline backs up.

Compare to `snykScan` and `updateSummaryPanel` which wrap in `CompletableFuture.runAsync` for the same reason.

**Fix:**
```java
@JsonNotification(value = LsConstants.SNYK_CONFIGURATION)
public void snykConfiguration(LspConfigurationParam param) {
    CompletableFuture.runAsync(() -> {
        try {
            if (param == null) return;
            // ... existing body ...
        } catch (Exception e) {
            SnykLogger.logError(e);
        }
    });
}
```

— AI review comment

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
```diff
@@ -375,4 +396,65 @@ public boolean isAuthenticated() {
...
+	public void clearExplicitlyChanged(String key) {
+		if (key == null) {
+			return;
+		}
+		explicitChanges.remove(key);
+		persistExplicitChanges();
+	}
+
+	private void persistExplicitChanges() {
+		insecurePreferences.put(EXPLICIT_CHANGES_KEY, String.join(",", explicitChanges));
```
acke, line 453: 🟡 **Medium**

**[Preferences.java:453]** `explicitChanges` is serialized as `String.join(",", explicitChanges)`. If any pref key ever contains a comma (current keys don't, but nothing enforces that), the round-trip via line 150 (`split(",")`) silently corrupts the set.

**Fix:**
Use a delimiter that can't appear in keys (e.g. `` Unit Separator), or escape commas. Cheaper: add a static assertion at registry-load time:
```java
for (LsSettingsRegistry.Entry e : LsSettingsRegistry.ENTRIES.values()) {
    if (e.prefKey \!= null && e.prefKey.indexOf(',') >= 0) {
        throw new IllegalStateException("pref key contains comma: " + e.prefKey);
    }
}
```

— AI review comment

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
```diff
@@ -24,74 +20,60 @@ public void configurationChanged() {
...
+			String rawValue;
+			if (LsKey.CLI_PATH == entry.lsKey) {
+				rawValue = preferences.getCliPath();
+			} else {
+				rawValue = preferences.getPref(entry.prefKey, entry.outboundDefault);
+			}
+
+			Object lsValue = entry.outboundSerializer.apply(rawValue);
+
+			boolean changed = entry.isAlwaysChanged || preferences.isExplicitlyChanged(entry.prefKey);
```
acke, line 47: 🟡 **Medium**

**[LsConfigurationUpdater.java:32-54]** `Entry.alwaysChanged` (TOKEN, TRUSTED_FOLDERS) sends the current pref value with `changed: true` on every config push. For TOKEN, when the user is unauthenticated, this sends `token: ""` with `changed: true` — which the LS may interpret as a deliberate logout request rather than "no token set yet". Worth confirming against LS behavior; if confirmed safe, add a one-line comment so a future reader doesn't reintroduce a regression by "fixing" it.

**Fix:**
Either (a) skip TOKEN outbound when value is empty AND user has never authenticated, or (b) add a comment at `Entry.encryptedAlwaysChanged` declaration in `LsSettingsRegistry.java:121` explaining why empty-token-with-changed=true is safe.

— AI review comment

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -409,21 +425,50 @@ public void snykScan(SnykScanParam param) {
...
+				var folderConfigSettings = FolderConfigSettings.getInstance();
+				folderConfigSettings.addAll(param.getFolderConfigs());
+				if (Preferences.getInstance().getBooleanPref(Preferences.SCANNING_MODE_AUTOMATIC)) {
+					triggerScan(null);
+				}
+				if (this.toolView != null) {
+					this.toolView.refreshDeltaReference();
+				}
+			}
+		} catch (Exception e) {
```
acke, line 441: 🔵 **Low**

**[SnykExtendedLanguageClient.java:449-451]** The `catch (Exception e)` in `snykConfiguration` swallows all failures into `SnykLogger.logError`. The LS receives no notification that its push was dropped — silent failure on the protocol surface.

**Fix:**
Acceptable to keep, but consider an `info`-level log at handler entry (`SnykLogger.logInfo("$/snyk.configuration: settings=N, folders=M")`) so operators can correlate inbound pushes with downstream state changes in logs.

— AI review comment

plugin/src/main/java/io/snyk/eclipse/plugin/html/ExecuteCommandBridge.java
```diff
@@ -111,9 +113,18 @@ static void saveLoginArgs(List<Object> args) {
...
     Preferences prefs = Preferences.getInstance();
-    prefs.store(Preferences.AUTHENTICATION_METHOD, authMethod);
-    prefs.store(Preferences.ENDPOINT_KEY, endpoint);
-    prefs.store(Preferences.INSECURE_KEY, insecure);
+    storeAndMarkIfChanged(prefs, LsKey.AUTHENTICATION_METHOD, authMethod);
+    storeAndMarkIfChanged(prefs, LsKey.ENDPOINT, endpoint);
+    storeAndMarkIfChanged(prefs, LsKey.INSECURE, insecure);
+    prefs.flushExplicitChanges();
+  }
+
```
acke, line 121: 🟡 **Medium**

**[ExecuteCommandBridge.java:125]** (Prior-comment follow-up.) Author deferred the revert-to-default semantics with "tracked as a known gap". Please leave a code-level marker so future maintainers can find it:
```java
// TODO(IDE-XXXX): reset-to-default does not yet clear explicitChanges; saving the registry
// default from the HTML form leaves changed=false on outbound sync. Implement when reset
// UX lands.
```
Without this marker the deferral lives only in the GitHub thread.

— AI review comment

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Dropped `INBOUND_ALLOWED_GLOBAL_KEYS` entirely. `persistGlobalSettings` now persists any key that has a `prefKey` in the registry (matching VS Code's `vscodeKey`-based filter), with one explicit skip for `LsKey.TOKEN` (auth credentials are IDE→LS only). Added `snykConfigurationNeverPersistsInboundToken` negative test confirming TOKEN is never overwritten by an inbound push.

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
nick-y-snyk: Won't fix. The migration only marks a pref explicit when its disk value ≠ registry default. A correct "reset all" implementation would write defaults back to disk, making the migration a no-op even if `explicitChanges` was cleared. The risk only materialises if a future reset clears `explicitChanges` without writing defaults to disk — a future implementation bug, not a current one.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Won't fix. `prefs.store()` calls `insecurePreferences.put()` which is in-memory only — Eclipse flushes `IEclipsePreferences` asynchronously via the platform. No synchronous disk I/O on the lsp4j dispatch thread. The only explicit `flush()` call is in `flushExplicitChanges()`, which `persistGlobalSettings` does not invoke.

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
nick-y-snyk: Won't fix. Current pref keys are all dot-separated identifiers — none contain commas, and the registry is closed (keys are defined by `LsKey` enum entries). A future key with a comma would be caught immediately in tests. Adding a runtime assertion would be defensive against a category of mistake that can't silently slip through.

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
nick-y-snyk: Won't fix. VS Code and IntelliJ both skip token outbound when empty, but the LS itself guards against this too: `refreshLdxSyncOnTokenChange` returns early when `newToken == ""` (configuration.go:230). Empty token with `changed=true` is a no-op on the LS side — no logout triggered.

plugin/src/main/java/io/snyk/eclipse/plugin/html/ExecuteCommandBridge.java
nick-y-snyk: Won't add. The deferral is documented in the PR description and GitHub thread. A TODO in code would reference a ticket that doesn't exist yet — when reset UX is scoped, the implementer will find this via the PR history or the `explicitChanges` / `storeAndMarkIfChanged` code paths.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Added `SnykLogger.logInfo("$/snyk.configuration received: settings=N, folders=M")` at handler entry — counts only, no payload, no token exposure. VS Code logs a similar debug-level entry; IntelliJ logs on settings-changed only. Eclipse has no debug level so this goes to info, but it's one line per notification.

acke (approved):
approved with comment.

PR #412, opened 2026-06-08
fix(tree-view): open file nodes in Eclipse editor instead of system default [IDE-2121]

nick-y-snyk:
### Description

Clicking a node in the HTML tree view (Secrets product) for files with no registered Eclipse editor (`.key`, `.md`, etc.) was opening the file in Windsurf (or whatever app macOS had set as default) instead of Eclipse.

Root cause: `showDocument` fell through to LSP4E's default handler which calls `IDE.openEditorOnFileStore` without an explicit editor ID — Eclipse's registry then returns "System Editor" for unknown types, delegating to macOS LaunchServices.

Fix: intercept `file://` URIs before they reach `super.showDocument()`, resolve the editor ID via `IDE.getEditorDescriptor` with external editors disallowed, and fall back to `"org.eclipse.ui.DefaultTextEditor"` — preventing any OS delegation.

Jira: https://snyksec.atlassian.net/browse/IDE-2121

### Checklist

- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

_Before: clicking a `.key` secret node → Windsurf opens._
_After: clicking a `.key` secret node → Eclipse default text editor opens with selection on the issue range._

### DIFF
--- a/plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
+++ b/plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
@@ -589,17 +589,17 @@ public class SnykExtendedLanguageClient extends LanguageClientImpl {
 	}
 
 	private IEditorPart openEditorForUri(IWorkbenchPage page, URI uri) throws PartInitException, CoreException {
-		uri = uri.normalize();
-		if (uri.getPath() != null && uri.getPath().contains("..")) {
-			SnykLogger.logInfo("Rejected URI with path traversal: " + uri);
+		URI normalizedUri = uri.normalize();
+		if (normalizedUri.getPath() != null && normalizedUri.getPath().contains("..")) {
+			SnykLogger.logInfo("Rejected URI with path traversal: " + normalizedUri);
 			return null;
 		}
-		String editorId = resolveInternalEditorId(uri);
-		IFile workspaceFile = LSPEclipseUtils.getFileHandle(uri.toString());
+		String editorId = resolveInternalEditorId(normalizedUri);
+		IFile workspaceFile = LSPEclipseUtils.getFileHandle(normalizedUri.toString());
 		if (workspaceFile != null && workspaceFile.exists()) {
 			return page.openEditor(new org.eclipse.ui.part.FileEditorInput(workspaceFile), editorId);
 		}
-		IFileStore fileStore = EFS.getStore(uri);
+		IFileStore fileStore = EFS.getStore(normalizedUri);
 		return page.openEditor(new FileStoreEditorInput(fileStore), editorId);
 	}
 
--- a/plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
+++ b/plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
@@ -532,8 +532,12 @@ public class SnykExtendedLanguageClient extends LanguageClientImpl {
 			issue = getIssueFromCache(uriDetails.product(), uriDetails.issueId());
 		} else if (FILE_SCHEME.equals(uriDetails.scheme())) {
 			return openFileInEclipse(uri, params.getSelection());
-		} else {
+		} else if ("http".equals(uriDetails.scheme()) || "https".equals(uriDetails.scheme())) {
 			return super.showDocument(params);
+		} else {
+			SnykLogger.logInfo(String.format("Unsupported URI: scheme=%s, product=%s, action=%s, issue=%s",
+					uriDetails.scheme(), uriDetails.product(), uriDetails.action(), uriDetails.issueId()));
+			return CompletableFuture.completedFuture(new ShowDocumentResult(false));
 		}
 
 		if (issue == null) {
--- a/plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
+++ b/plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
@@ -532,12 +532,8 @@ public class SnykExtendedLanguageClient extends LanguageClientImpl {
 			issue = getIssueFromCache(uriDetails.product(), uriDetails.issueId());
 		} else if (FILE_SCHEME.equals(uriDetails.scheme())) {
 			return openFileInEclipse(uri, params.getSelection());
-		} else if ("http".equals(uriDetails.scheme()) || "https".equals(uriDetails.scheme())) {
-			return super.showDocument(params);
 		} else {
-			SnykLogger.logInfo(String.format("Unsupported URI: scheme=%s, product=%s, action=%s, issue=%s",
-					uriDetails.scheme(), uriDetails.product(), uriDetails.action(), uriDetails.issueId()));
-			return CompletableFuture.completedFuture(new ShowDocumentResult(false));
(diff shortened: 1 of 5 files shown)


plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -540,6 +554,72 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
...
+	// unknown file types to avoid routing through macOS LaunchServices (e.g. Windsurf).
+	private CompletableFuture<ShowDocumentResult> openFileInEclipse(URI uri, Range range) {
+		if (Preferences.getInstance().isTest()) {
+			return CompletableFuture.completedFuture(new ShowDocumentResult(false));
+		}
+		return CompletableFuture.supplyAsync(() -> {
+			boolean[] success = { false };
+			Display.getDefault().syncExec(() -> {
+				try {
+					IWorkbenchPage page = PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage();
```
bastiandoetsch: **Should Fix: `getActiveWorkbenchWindow()` can return null — NPE inside `syncExec`**

`getActiveWorkbenchWindow()` returns `null` when Eclipse has no active window (minimized, during startup, all windows closed). Chaining directly to `.getActivePage()` produces an NPE, which is caught by the outer `catch(Exception e)` and logged — but returns `success[0]=false` with no diagnostic context.

The line 569 `page == null` guard only fires when the window is non-null but has no active page.

```java
IWorkbenchWindow window = PlatformUI.getWorkbench().getActiveWorkbenchWindow();
if (window == null) return;
IWorkbenchPage page = window.getActivePage();
```
— AI review

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -540,6 +554,72 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
...
+				return descriptor.getId();
+			}
+		} catch (PartInitException e) {
+			SnykLogger.logError(e);
+		}
+		return "org.eclipse.ui.DefaultTextEditor";
+	}
+
+	private void revealRange(ITextEditor textEditor, Range range) {
+		IDocument doc = textEditor.getDocumentProvider().getDocument(textEditor.getEditorInput());
```
bastiandoetsch: **Should Fix: `getDocumentProvider()` can return null — NPE before null check on `doc`**

`getDocumentProvider()` may return null for a disposed or not-fully-initialized editor. The null check at line 611 is on `doc`, not on the provider itself — so if the provider is null, an NPE is thrown inside `syncExec`, `success[0]` stays at the value set before `revealRange` was called (i.e., `true`), and the caller receives `ShowDocumentResult(true)` while the range reveal silently failed.

```java
IDocumentProvider provider = textEditor.getDocumentProvider();
if (provider == null) return;
IDocument doc = provider.getDocument(textEditor.getEditorInput());
```

Also consider moving `success[0] = true` to after `revealRange` completes so partial success (file opened, position not set) is not reported as full success. — AI review

bastiandoetsch:
## AI Review: snyk-eclipse-plugin #412 — fix(tree-view): open file nodes in Eclipse editor [IDE-2121]

**Verdict: approved with Should-Fix items (see inline comments)**

> Adversarial agent voted REJECT (threading deadlock, isTest coverage gaps); semantic agent verified the NPEs are caught by outer `catch(Exception e)` and threading is correct. Ruling: the NPEs are real Should-Fix bugs (silent partial failure), but with outer catch the function doesn't crash — approving with the fixes flagged.

### Should Fix (in-diff)
- `SnykExtendedLanguageClient.java:567` — `getActiveWorkbenchWindow()` not null-checked before chaining `.getActivePage()`. NPE caught by outer catch but reports misleading `success=false` with no context. (See inline comment.)
- `SnykExtendedLanguageClient.java:610` — `getDocumentProvider()` not null-checked. NPE caught by outer catch but `success[0]` was already set `true`, so caller gets `ShowDocumentResult(true)` while range reveal silently failed. (See inline comment.)
- `SnykExtendedLanguageClient.java:593` — `EFS.getStore(uri)` receives un-normalized URI. A crafted LS response with `../../etc/shadow` path traversal would open that file in the text editor. Add `uri = uri.normalize()` and reject paths containing `..` at entry point of `openEditorForUri`.
- `plugin/pom.xml:69` — `commons-logging:1.3.4` added with no visible usage in diff. Remove if not intentional, or add comment explaining why (CVE pin?).

### Should Fix (out-of-diff)
- Tests only exercise the `isTest()` early-exit guard — the actual editor-open routing is untested. Consider extracting `resolveInternalEditorId` and `revealRange` for standalone unit tests, or adding a mock-based test that verifies `file://` URIs do NOT reach `super.showDocument`.
- `resolveInternalEditorId:599` — fallback to `DefaultTextEditor` for files with no registered internal editor will open binary files (images, PDFs, `.class`) as garbage text. Return `null` instead, letting `openEditorForUri` no-op for unrecognized types.

### Suggestions
- `getEditorDescriptor(path, allowInteractive=true, ...)` — for completely unregistered file types, Eclipse shows an "Open With" dialog before throwing `PartInitException`. Use `false` to suppress this dialog; `DefaultTextEditor` fallback still applies.
- `SnykExtendedLanguageClient` now owns EFS stores, editor selection, and range reveal — consider extracting to `EclipseEditorOpener` utility for testability.

### Verified Correct
- Threading: `Display.syncExec` inside `supplyAsync` is correct for returning a result. `boolean[] success = {false}` array is safe because `syncExec` blocks until UI work completes. No threading rule violated per semantic analysis.
- `openEditorForUri` dual-path (workspace `IFile` vs EFS `IFileStore`) is correct Eclipse pattern for both workspace and external files.
- `resolveInternalEditorId` `!descriptor.isOpenExternal()` check is the key fix preventing OS delegation (Windsurf/macOS LaunchServices intercept). Correct.
- URI scheme routing (file/snyk/other) is clean and correctly structured.
- Snyk SCA/Code: 0 new issues. Pre-existing low-severity code findings not introduced by this PR. — AI review


[IDE-2121]: https://snyksec.atlassian.net/browse/IDE-2121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

bastiandoetsch (approved):
No confirmed crash paths (outer catch covers NPEs). Should-Fix: null-check getActiveWorkbenchWindow/getDocumentProvider (inline), URI normalization, spurious commons-logging. — AI review

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Fixed: split getActiveWorkbenchWindow().getActivePage() into two steps with null guards for both window and page.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Fixed: extracted getDocumentProvider() into a local variable with null guard before calling getDocument().

PR #413, opened 2026-06-09
fix(tree-view): show issue details panel when clicking HTML tree nodes [IDE-2124]

nick-y-snyk:
### Description

Fixes [IDE-2124](https://snyksec.atlassian.net/browse/IDE-2124).

When `USE_HTML_TREE_VIEW` is enabled (default since IDE-1837), clicking an issue node in the HTML tree showed the default "Select an issue and start improving your project." placeholder instead of issue details.

**Root cause:** `selectTreeNode()` called `treeViewer.setSelection()` on the hidden `StackLayout` control — `TreeViewer` is behind the HTML tree `Browser`. SWT does not fire `ISelectionChangedListener` for `setSelection()` on a non-visible control, so `browserHandler.updateBrowserContent()` was never reached.

**Fix:** When HTML tree mode is active, find the `IssueTreeNode` directly and call `browserHandler.updateBrowserContent()` without routing through the hidden SWT tree's selection listener.

### Checklist

- [ ] Read and understood the [Code of Conduct](https://github.com/snyk/snyk-eclipse-plugin/blob/main/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/snyk/snyk-eclipse-plugin/blob/main/CONTRIBUTING.md).
- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

_Before: details panel shows placeholder on HTML tree node click. After: issue details render correctly._

[IDE-2124]: https://snyksec.atlassian.net/browse/IDE-2124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
+++ b/plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
@@ -555,8 +555,11 @@ public class SnykToolView extends ViewPart implements ISnykToolView {
 			if (productNode != null) {
 				IssueTreeNode issueNode = findIssueTreeNode((TreeNode) productNode, issue.id());
 				if (issueNode != null) {
-					selectedNode = issueNode;
-					Display.getDefault().asyncExec(() -> browserHandler.updateBrowserContent(issueNode));
+					final IssueTreeNode finalNode = issueNode;
+					runOnDisplay(() -> {
+						selectedNode = finalNode;
+						browserHandler.updateBrowserContent(finalNode);
+					});
 				}
 			}
 			return;
@@ -580,18 +583,12 @@ public class SnykToolView extends ViewPart implements ISnykToolView {
 	}
 
 	private void selectTreenodeForIssue(TreeNode currentParent, Issue issue) {
-		for (Object child : currentParent.getChildren()) {
-			TreeNode childNode = (TreeNode) child;
+		IssueTreeNode node = findIssueTreeNode(currentParent, issue.id());
+		if (node != null) updateSelection(node);
+	}
 
-			if (childNode instanceof IssueTreeNode && ((IssueTreeNode) childNode).getIssue().id().equals(issue.id())) {
-				updateSelection((IssueTreeNode) childNode);
-				return; // Exit the function as we've found a match
-			}
-
-			if (childNode.getChildren() != null && childNode.getChildren().length != 0) {
-				selectTreenodeForIssue(childNode, issue);
-			}
-		}
+	protected void runOnDisplay(Runnable runnable) {
+		Display.getDefault().asyncExec(runnable);
 	}
 
 	private void updateSelection(IssueTreeNode issueTreeNode) {
--- a/plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
+++ b/plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
@@ -548,12 +548,15 @@ public class SnykToolView extends ViewPart implements ISnykToolView {
 			if (treeBrowserHandler != null) {
 				Display.getDefault().asyncExec(() -> treeBrowserHandler.selectNode(issue.id()));
 			}
-			// selectTreenodeForIssue triggers selectionChanged on the hidden SWT tree,
-			// which updates the detail browser panel via browserHandler.updateBrowserContent.
-			if (treeViewer != null) {
-				ProductTreeNode productNode = getProductNode(product, issue.filePath());
-				if (productNode != null) {
-					selectTreenodeForIssue((TreeNode) productNode, issue);
+			// treeViewer.setSelection on a hidden StackLayout control does not fire
+			// ISelectionChangedListener, so drive the detail panel update directly.
+			if (treeViewer == null) return;
+			ProductTreeNode productNode = getProductNode(product, issue.filePath());
+			if (productNode != null) {
+				IssueTreeNode issueNode = findIssueTreeNode((TreeNode) productNode, issue.id());
+				if (issueNode != null) {
+					selectedNode = issueNode;
+					Display.getDefault().asyncExec(() -> browserHandler.updateBrowserContent(issueNode));
 				}
 			}
 			return;
@@ -561,6 +564,19 @@ public class SnykToolView extends ViewPart implements ISnykToolView {
 		ProductTreeNode productNode = getProductNode(product, issue.filePath());
 		if (productNode == null) return;
 		selectTreenodeForIssue((TreeNode) productNode, issue);
+	}
+
+	private IssueTreeNode findIssueTreeNode(TreeNode parent, String issueId) {
+		if (parent == null || parent.getChildren() == null) return null;
+		for (Object child : parent.getChildren()) {
+			TreeNode node = (TreeNode) child;
+			if (node instanceof IssueTreeNode && ((IssueTreeNode) node).getIssue().id().equals(issueId)) {
+				return (IssueTreeNode) node;
+			}
+			IssueTreeNode found = findIssueTreeNode(node, issueId);
+			if (found != null) return found;
+		}
+		return null;
 	}
 
 	private void selectTreenodeForIssue(TreeNode currentParent, Issue issue) {
--- a/tests/src/test/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolViewTest.java
+++ b/tests/src/test/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolViewTest.java
@@ -2,10 +2,14 @@ package io.snyk.eclipse.plugin.views.snyktoolview;
 
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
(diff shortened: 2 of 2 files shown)


acke (approved):
## Orchestrated review (review-orchestrator)

**Verdict: APPROVE WITH NITS** — root-cause analysis is correct (SWT does not fire `ISelectionChangedListener` on a control hidden behind `StackLayout`) and the fix is minimal and targeted. Two nits worth landing before merge (off-UI-thread write to `selectedNode`, missing behavioral test for the new HTML path) and one optional cleanup (duplicate tree traversal). Inline comments cover each.

---
🤖 review-orchestrator (1 file, +22/−6 — skipped full 5-lens fan-out for size)

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
```diff
@@ -548,12 +548,15 @@ public void selectTreeNode(Issue issue, String product) {
...
-				if (productNode != null) {
-					selectTreenodeForIssue((TreeNode) productNode, issue);
+			// treeViewer.setSelection on a hidden StackLayout control does not fire
+			// ISelectionChangedListener, so drive the detail panel update directly.
+			if (treeViewer == null) return;
+			ProductTreeNode productNode = getProductNode(product, issue.filePath());
+			if (productNode != null) {
+				IssueTreeNode issueNode = findIssueTreeNode((TreeNode) productNode, issue.id());
+				if (issueNode != null) {
+					selectedNode = issueNode;
```
acke: 🟡 **Medium**

**[SnykToolView.java:558]** `selectedNode = issueNode;` runs on the caller's thread. `selectTreeNode()` is invoked from `CompletableFuture.supplyAsync(...)` in `SnykExtendedLanguageClient.showDocument` (line 530) — a worker thread. Every other write to `selectedNode` happens on the UI thread inside `Display.asyncExec` (e.g. the selection listener at line 152). UI-thread readers (context-menu actions at lines 202, 247, 283) can now race on this non-volatile field with no happens-before. The asyncExec one line below already does the right thing for `browserHandler.updateBrowserContent` — just lift the assignment into it.

**Fix:**

```java
final IssueTreeNode finalNode = issueNode;
Display.getDefault().asyncExec(() -> {
    selectedNode = finalNode;
    browserHandler.updateBrowserContent(finalNode);
});
```

— AI review comment

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
```diff
@@ -563,6 +566,19 @@ public void selectTreeNode(Issue issue, String product) {
 		selectTreenodeForIssue((TreeNode) productNode, issue);
 	}
 
+	private IssueTreeNode findIssueTreeNode(TreeNode parent, String issueId) {
```
acke, line 572: 🔵 **Low**

**[SnykToolView.java:569-580]** `findIssueTreeNode` duplicates the traversal already in `selectTreenodeForIssue` (lines 582-595). The new helper is actually the better-behaved variant — it returns the match and short-circuits. `selectTreenodeForIssue` has a latent bug where, after recursing into a child that contained a match, the outer loop keeps iterating siblings and can call `updateSelection` more than once. Collapsing the two removes the duplication and fixes the latent issue.

**Fix:**

```java
private void selectTreenodeForIssue(TreeNode currentParent, Issue issue) {
    IssueTreeNode node = findIssueTreeNode(currentParent, issue.id());
    if (node \!= null) updateSelection(node);
}
```

— AI review comment

acke:
🟡 **Medium**

**[PR checklist / tests/]** The PR checklist has "Tests added and all succeed" ticked, but the diff is one production file — no test in this PR covers the new behavior. The existing `SnykToolViewTest` only asserts `assertDoesNotThrow` in HTML-mode + null-handler paths; nothing verifies the actual fix — that `browserHandler.updateBrowserContent(issueNode)` fires when an issue is selected and the productNode resolves. This is the exact path that regressed; without a test, the next refactor can re-break it silently.

**Fix:**
Add a behavioral test in `SnykToolViewTest`: inject a `BrowserHandler` mock (via the same reflection seam used for `pendingHtml`), seed a `ProductTreeNode` containing an `IssueTreeNode` with a matching id, enable `USE_HTML_TREE_VIEW`, call `selectTreeNode(issue, product)`, and `verify(browserHandler).updateBrowserContent(...)`.

— AI review comment

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
nick-y-snyk: Fixed — moved `selectedNode = issueNode` into the `runOnDisplay` (asyncExec) lambda so both the assignment and the browser update happen on the UI thread. Introduced a `protected runOnDisplay(Runnable)` seam to keep it testable without a live SWT Display.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
nick-y-snyk: Fixed — collapsed `selectTreenodeForIssue` to delegate to `findIssueTreeNode`, removing the duplicate traversal and the latent multi-call bug.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
```diff
@@ -548,12 +548,18 @@ public void selectTreeNode(Issue issue, String product) {
...
+			// ISelectionChangedListener, so drive the detail panel update directly.
+			if (treeViewer == null) return;
+			ProductTreeNode productNode = getProductNode(product, issue.filePath());
+			if (productNode != null) {
+				IssueTreeNode issueNode = findIssueTreeNode((TreeNode) productNode, issue.id());
+				if (issueNode != null) {
+					final IssueTreeNode finalNode = issueNode;
+					runOnDisplay(() -> {
+						selectedNode = finalNode;
+						browserHandler.updateBrowserContent(finalNode);
```
bastiandoetsch, line 561: **Should Fix:** `treeBrowserHandler` is null-checked before use (line 548), but `browserHandler` here is not. Both are assigned in `createPartControl`, so in normal lifecycle they're always non-null together — but the guard on `treeViewer != null` at line 553 doesn't transitively guarantee `browserHandler` is set (e.g. if `createPartControl` is interrupted, or in tests where `treeViewer` is injected via reflection before `browserHandler`). Add `if (browserHandler == null) return;` before this `runOnDisplay` block, mirroring the `treeBrowserHandler` guard pattern already present above. — AI review

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
```diff
@@ -563,19 +569,26 @@ public void selectTreeNode(Issue issue, String product) {
...
 		}
+		return null;
+	}
+
+	private void selectTreenodeForIssue(TreeNode currentParent, Issue issue) {
+		IssueTreeNode node = findIssueTreeNode(currentParent, issue.id());
+		if (node != null) updateSelection(node);
+	}
+
+	protected void runOnDisplay(Runnable runnable) {
```
bastiandoetsch, line 590: **Suggestion:** Making `runOnDisplay` `protected` exposes a test seam as a subclassing extension point. A subclass overriding this to run synchronously on a non-UI thread would cause `SWTException: Invalid thread access` on all `Display`/`Browser` calls inside the lambda. Prefer making it `/* package-private */` (no modifier) or injecting a `Runnable` executor via a package-private setter used only by tests. — AI review

bastiandoetsch:
## AI Review Summary — snyk/snyk-eclipse-plugin #413

**Verdict: Changes Requested** — 1 Should Fix (inline), 1 Suggestion (inline), plus notes below.

### Should Fix
1. **`browserHandler` missing null-check** (`SnykToolView.java:561`) — `treeBrowserHandler` is null-guarded (line 548) but `browserHandler` is not before `browserHandler.updateBrowserContent(finalNode)`. Both are set in `createPartControl`, but the `treeViewer != null` guard at line 553 doesn't transitively guarantee `browserHandler` is set (interrupted init, test injection via reflection). Add a matching null-check. (inline comment posted)

2. **Threading concerns** (noted by acke's review): `findIssueTreeNode` and `selectedNode` writes — supporting acke's existing Medium findings. `selectedNode` should document its SWT-thread-confinement assumption; `findIssueTreeNode` should be called from inside the `runOnDisplay` lambda or from the UI thread only.

3. **CHANGELOG.md not updated** — PR checklist requires it; this is a user-facing bug fix (HTML tree is the default since IDE-1837).

### Suggestion
4. **`runOnDisplay` as `protected`** (`SnykToolView.java:590`) — exposes a test seam as a subclassing extension point; a subclass running this synchronously on a non-UI thread would cause `SWTException`. Prefer package-private or constructor-injected executor. (inline comment posted)

### Verified Correct
- **File-open not regressed** — `LSPEclipseUtils.open()` was already unreachable in HTML tree mode before this PR (`ISelectionChangedListener` never fired on the hidden StackLayout control). The new direct path is not a regression.
- **`findIssueTreeNode` return propagation** — correctly returns the first match and propagates through recursion. Old code discarded the recursive call's result; this is a real improvement.
- **CI clean** — Snyk scans: 0 issues in changed files.

— AI review

PR #416, opened 2026-06-10
refactor: drop SWT tree, SnykIssueCache, and call getIssueDescription directly [IDE-2126]

nick-y-snyk:
### Description

Drops the entire cache layer that existed only to back the deleted SWT tree view.

- Deletes `SnykIssueCache`, `IssueCacheHolder`, `ScanState`, `ScanInProgressKey`, `Diagnostic316`, `PublishDiagnostics316Param` (~980 lines)
- `showDocument` now calls `getIssueDescription(issueId)` directly via the LS `generateIssueDescription` command — no cache lookup
- `BrowserHandler.updateBrowserContent` takes `String issueId` instead of `Issue`
- `ISnykToolView.selectTreeNode` takes `String issueId` instead of `Issue + product`
- `ClearCacheHandler` and `StopScanHandler` stripped of dead `ScanState` calls
- `ProductConstants.LSP_SOURCE_TO_SCAN_PARAMS` removed (only fed the cache)
- `build.properties` corrected: `commons-logging-1.2.jar` (was `1.3.4`)
- Tests updated to match new signatures; cache-backed showDocument tests replaced with direct mock-toolview tests

Jira: https://snyksec.atlassian.net/browse/IDE-2126

### Checklist

- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

Pre-LS init:
Screenshot 2026-06-12 at 15 42 03
White box is not a regression.

Pre-auth:
Screenshot 2026-06-12 at 15 42 18

After scan selecting an issue:
Screenshot 2026-06-12 at 15 43 18


plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/handlers/ClearCacheHandler.java
rrama, line 1: Can this not be deleted (along with the button)?
If so, do as a follow up PR (no point doing in this PR).

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/handlers/CollapseTreeHandler.java
rrama, line 1: Can this not be deleted (along with the button)?
If so, do as a follow up PR (no point doing in this PR).

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/handlers/FilterFixableIssuesHandler.java
rrama, line 1: Again, can be deleted (as long as the uses are correctly updated, as this one does call `super.execute`).
If so, do as a follow up PR (no point doing in this PR).

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/handlers/StopScanHandler.java
rrama, line 1: Q: Do we let LS know the scan has been stopped?

plugin/META-INF/MANIFEST.MF
```diff
@@ -27,7 +27,9 @@ Require-Bundle: org.eclipse.ui,
  org.eclipse.ui.editors,
  org.eclipse.ui.genericeditor,
  org.eclipse.swt,
- org.eclipse.jface
+ org.eclipse.jface,
+ org.eclipse.core.filesystem,
+ org.eclipse.ui.ide
```
rrama: Undo these changes, they already exist.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/handlers/BaseSeverityFilterHandler.java
rrama, line 1: Again, can be deleted (as long as the uses are correctly updated, as this one does call `super.execute`).
If so, do as a follow up PR (no point doing in this PR).

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BrowserHandler.java
```diff
@@ -164,70 +160,19 @@ public void completed(ProgressEvent event) {
...
 			if (isEmpty(htmlContent)) {
 				htmlContent = htmlProvider.getNoDescriptionHtml();
 			}
-
 			final var browserContent = htmlProvider.replaceCssVariables(htmlContent);
-
-			Display.getDefault().syncExec(() -> {
-				browser.setText(browserContent);
-			});
+			Display.getDefault().syncExec(() -> browser.setText(browserContent));
```
rrama: Needs a disposed check.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/ISnykToolView.java
```diff
@@ -23,64 +19,6 @@ public interface ISnykToolView {
 	public String NODE_TEXT_SCANNING = "Scanning...";
 	public String NODE_TEXT_ERROR = "(error)";
```
rrama: Are these needed anymore?
If not, do as a follow up PR (no point doing in this PR).

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BrowserHandler.java
```diff
@@ -164,70 +160,19 @@ public void completed(ProgressEvent event) {
...
 
 
-	public CompletableFuture<Void> updateBrowserContent(TreeNode node) {
-		// Generate HTML content based on the selected node
-		var htmlProvider = getHtmlProvider(node);
-		initScript = htmlProvider.getInitScript();
-		boolean shouldShowDefaultMessage = true;
-		if (node instanceof ProductTreeNode) {
-			var ptn = (ProductTreeNode) node;
-			var presentableError = ptn.getPresentableError();
```
rrama, line 174: Q: What happens to scan errors do they no longer show up in the details panel?
Do we do anything with the presentable errors anymore?

tests/src/test/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolViewTest.java
```diff
@@ -77,73 +56,10 @@ void updateTreeViewHtml_multipleCallsWhenHandlerNull_keepsLatestHtml() throws Ex
...
-	@Test
-	void selectTreeNode_withHtmlTreeViewEnabled_andNullIssue_doesNotThrow() {
-		Preferences.getTestInstance(new InMemoryPreferenceStore(), new InMemorySecurePreferenceStore())
-				.store(Preferences.USE_HTML_TREE_VIEW, "true");
-		SnykToolView view = new SnykToolView();
-		assertDoesNotThrow(() -> view.selectTreeNode(null, "Snyk Code"));
-	}
-
-	@Test
-	void selectTreeNode_withHtmlTreeViewEnabled_updatesContentPanel() throws Exception {
```
rrama, line 101: You don't have a test which replaces this one to prove the content actually tried to render. You will need to add some new mocks to actually be able to test it.

AI had this to say:

> The old selectTreeNode_withHtmlTreeViewEnabled_updatesContentPanel test verified updateBrowserContent was actually invoked, but none of the surviving tests assert that selecting a node dispatches to the browser. In headless tests Display.getDefault() throws SWTError, which selectTreeNode catches and swallows (SnykToolView.java:155-165), so these assertDoesNotThrow cases short-circuit before reaching browserHandler.updateBrowserContent(issueId) and never exercise the new HTML path. Can we add a test that injects mock browserHandler/treeBrowserHandler, runs the asyncExec lambda synchronously, and verifys updateBrowserContent("issue-id") (and selectNode("issue-id")) is called? That would also cover the missing browserHandler null guard.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
```diff
@@ -67,317 +33,46 @@ public SnykToolView() {
...
+			treeBrowserHandler.setBrowserText(buffered);
 		}
-		stackLayout.topControl = useHtmlTreeView ? treeBrowser : treeViewer.getControl();
-		treeStack.layout();
-		// verticalSashForm has exactly 2 children: summaryBrowser and treeStack.
+
 		verticalSashForm.setWeights(1, 3);
 
-		// Create Browser
-		// SWT.EDGE will be ignored if OS not windows and will be set to SWT.NONE.
```
rrama, line 135: This is a useful comment and should not have been deleted.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BrowserHandler.java
```diff
@@ -164,70 +160,19 @@ public void completed(ProgressEvent event) {
...
-		}
-
-		return CompletableFuture.supplyAsync(() -> {
-			return generateHtmlContent(node);
-		}).thenAccept(htmlContent -> {
+	public void updateBrowserContent(String issueId) {
+		CompletableFuture.runAsync(() -> {
+			BaseHtmlProvider htmlProvider = new BaseHtmlProvider();
+			initScript = htmlProvider.getInitScript();
+			String htmlContent = SnykExtendedLanguageClient.getInstance().getIssueDescription(issueId);
```
rrama, line 173: Q: Have you tried spam clicking the tree nodes? I am wondering how quickly you would have to click from one to another for a race condition. If you can't produce a race condition, then all good and we can leave this as-is with no async order guarding.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
```diff
@@ -542,67 +152,22 @@ protected void outputCommandResult(Object result) {
...
-	private IssueTreeNode findIssueTreeNode(TreeNode parent, String issueId) {
-		if (parent == null || parent.getChildren() == null) return null;
-		for (Object child : parent.getChildren()) {
-			TreeNode node = (TreeNode) child;
-			if (node instanceof IssueTreeNode && ((IssueTreeNode) node).getIssue().id().equals(issueId)) {
-				return (IssueTreeNode) node;
-			}
-			IssueTreeNode found = findIssueTreeNode(node, issueId);
-			if (found != null) return found;
+			display.asyncExec(() -> browserHandler.updateBrowserContent(issueId));
```
rrama: Maybe should add a `null` guard like above.

tests/src/test/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolViewTest.java
```diff
@@ -77,73 +56,10 @@ void updateTreeViewHtml_multipleCallsWhenHandlerNull_keepsLatestHtml() throws Ex
...
-
-		// Inject mock treeViewer so the treeViewer == null guard does not short-circuit.
-		Field treeViewerField = SnykToolView.class.getDeclaredField("treeViewer");
-		treeViewerField.setAccessible(true);
-		treeViewerField.set(view, mockTreeViewer);
-
-		view.selectTreeNode(issue, "Snyk Code");
-
-		verify(mockBrowserHandler).updateBrowserContent(issueNode);
+		assertDoesNotThrow(() -> view.selectTreeNode("some-issue-id"));
```
rrama: This test and possibly others only pass because the exception (from the test being headless) is swallowed. Can we do better to increase the actual prod path coverage?

plugin/src/main/java/io/snyk/eclipse/plugin/domain/ProductConstants.java
```diff
@@ -29,16 +31,11 @@ public final class ProductConstants {
...
-			FILTERABLE_ISSUE_INFRASTRUCTURE_AS_CODE, DISPLAYED_IAC, 
+			FILTERABLE_ISSUE_INFRASTRUCTURE_AS_CODE, DISPLAYED_IAC,
 			FILTERABLE_ISSUE_OPEN_SOURCE, DISPLAYED_OSS);
 
-	public static final Map<String, String> LSP_SOURCE_TO_SCAN_PARAMS = Map.of(
-			DIAGNOSTIC_SOURCE_SNYK_CODE, SCAN_PARAMS_CODE, 
-			DIAGNOSTIC_SOURCE_SNYK_IAC, SCAN_PARAMS_IAC, 
-			DIAGNOSTIC_SOURCE_SNYK_OSS, SCAN_PARAMS_OSS);
-
 	public static final Map<String, String> SCAN_PARAMS_TO_DISPLAYED = Map.of(
```
rrama: Is this still needed now?
If not, do as a follow up PR (no point doing in this PR).

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
```diff
@@ -542,67 +152,22 @@ protected void outputCommandResult(Object result) {
...
+		} catch (SWTError | SWTException | UnsatisfiedLinkError | NoClassDefFoundError e) {
+			SnykLogger.logInfo("No SWT Display available for selectTreeNode: " + e.getMessage());
```
rrama, line 131: I see the intent of the `catch`, but I am not sure I agree it is the right thing to do. Let's discuss please.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
```diff
@@ -451,71 +104,28 @@ public void toggleIgnoresButtons() {
...
-		for (BaseTreeNode node : children) {
-			if (node instanceof ContentRootNode) {
-				ContentRootNode contentNode = (ContentRootNode) node;
-				if (!contentNode.getLabel().isBlank()) {
-					setReferenceText(contentNode);
-				}
-			}
-		}
-		refreshTree();
+		// Delta reference display handled by HTML tree view
```
rrama, line 104: Then can we just delete this function? Same with the ones below.
If so, do as a follow up PR (no point doing in this PR).

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
```diff
@@ -542,67 +152,22 @@ protected void outputCommandResult(Object result) {
...
-			IStructuredSelection selection = new StructuredSelection(issueTreeNode);
-			treeViewer.setSelection(selection, true);
-		});
 	}
 
 	@Override
 	public void updateTreeViewHtml(String html) {
-		// Always store first (last-write-wins), then drain on the UI thread.
-		// Since createPartControl also runs on the UI thread, the drain callbacks
-		// are serialized — whichever runs last when treeBrowserHandler is non-null wins.
```
rrama, line 605: Is this comment still useful and needed?

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -527,31 +449,22 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
...
 		if (uriDetails.isValid()) {
-			issue = getIssueFromCache(uriDetails.product(), uriDetails.issueId());
+			String issueId = uriDetails.issueId();
+			return CompletableFuture.supplyAsync(() -> {
+				openToolView();
+				ISnykToolView view = this.toolView;
+				if (view == null) {
+					return new ShowDocumentResult(false);
+				}
+				view.selectTreeNode(issueId);
```
rrama: Very minor possible nitpick: Should probably make this return a boolean (for if it is valid or not) which then is fed into `new ShowDocumentResult(result)`. Although I must admit I haven't looked into what the result is used for.
Edit: Giving it more thought, is it just used to tell Eclipse that we handled the URL and it doesn't need to do anything more?

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -1,41 +1,19 @@
...
-import static io.snyk.eclipse.plugin.domain.ProductConstants.DISPLAYED_CODE_SECURITY;
-import static io.snyk.eclipse.plugin.domain.ProductConstants.FILTERABLE_ISSUE_TYPE_TO_DISPLAY;
-import static io.snyk.eclipse.plugin.domain.ProductConstants.LSP_SOURCE_TO_SCAN_PARAMS;
-import static io.snyk.eclipse.plugin.domain.ProductConstants.SCAN_PARAMS_CODE;
-import static io.snyk.eclipse.plugin.domain.ProductConstants.SCAN_PARAMS_IAC;
-import static io.snyk.eclipse.plugin.domain.ProductConstants.SCAN_PARAMS_OSS;
-import static io.snyk.eclipse.plugin.domain.ProductConstants.SCAN_PARAMS_TO_DISPLAYED;
 import static io.snyk.eclipse.plugin.domain.ProductConstants.SCAN_STATE_ERROR;
 import static io.snyk.eclipse.plugin.domain.ProductConstants.SCAN_STATE_IN_PROGRESS;
 import static io.snyk.eclipse.plugin.domain.ProductConstants.SCAN_STATE_SUCCESS;
```
rrama, line 5: Please re-do imports, these are also now unused.
I'm surprised we don't have a PMD for it.

rrama (commented):
I don't think this is tested properly and seems to be a half-done removal, so I cannot approve it. An example, my severity filter changes in the native drop-down now does nothing. TBH, these can probably be removed, but they shouldn't just be unwired and left.
Image

> ### Screenshots / GIFs
> No UI change.

There are defo UI changes here! The HTML rendering for issues is redone. Maybe a screen recording would be nice to show it working.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BrowserHandler.java
rrama: Oh, I tried clicking quickly between nodes and it was fine, so all good. :)

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/handlers/ClearCacheHandler.java
nick-y-snyk: Done — deleted along with all other filter handler classes in this cleanup.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/handlers/CollapseTreeHandler.java
nick-y-snyk: Done — deleted along with all other filter handler classes in this cleanup.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/handlers/FilterFixableIssuesHandler.java
nick-y-snyk: Done — deleted along with all other filter handler classes in this cleanup.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/handlers/BaseSeverityFilterHandler.java
nick-y-snyk: Done — deleted along with all other filter handler classes in this cleanup.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
nick-y-snyk: Done — deleted along with all other filter handler classes in this cleanup.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BrowserHandler.java
nick-y-snyk: Done — added `!browser.isDisposed()` guard in the `syncExec` lambda.

tests/src/test/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolViewTest.java
nick-y-snyk: Done — extracted `dispatchTreeNode` and `dispatchBrowserContent` package-private methods from `selectTreeNode` so tests can inject mock `BrowserHandler`/`TreeViewBrowserHandler` via reflection and verify the calls without needing a real SWT Display. Five new tests added covering both handlers present, tree handler null, and browser handler null.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Already addressed in earlier commits — all unused imports (DIAGNOSTIC_SOURCE_*, SCAN_PARAMS_*, SEVERITY_*, tree node types, SnykIssueCache, etc.) were removed. Current import list is clean.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
nick-y-snyk: The catch is there because `Display.getDefault()` throws `SWTError` / `NoClassDefFoundError` in headless environments (tests, early startup before SWT is initialised). Without it, an LS notification arriving before the view is fully initialised would crash the plugin thread silently. Same defensive pattern used in `drainPendingHtmlAsync`. Happy to change the approach if you have a preferred pattern — e.g. checking `PlatformUI.isWorkbenchRunning()` first instead of catching.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
nick-y-snyk: Restored — adapted to the new HTML tree context and placed on `dispatchBrowserContent` where the direct dispatch now lives.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
nick-y-snyk: Already present — `drainPendingHtmlAsync` checks `display != null && !display.isDisposed()` before calling `asyncExec` (line 153).

tests/src/test/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolViewTest.java
nick-y-snyk: Addressed by the new `dispatchTreeNode`/`dispatchBrowserContent` tests added in this PR — those inject mocks and exercise the actual prod path (updateBrowserContent + selectNode) without relying on SWT. The remaining `assertDoesNotThrow` tests on `selectTreeNode` now only cover the null/empty-id guard branches, which are trivially safe.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Yes, exactly — `ShowDocumentResult(boolean)` tells the LS client whether the IDE handled the URI. The method already returns `false` when the view is unavailable and `true` on success, so the boolean is already threaded through correctly. No change needed.

nick-y-snyk: Done — extracted `selectIssueNode(String)` returning `boolean`, fed into `new ShowDocumentResult(result)`. `showDocument` is now a one-liner for the valid-URI path.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BrowserHandler.java
nick-y-snyk: Scan errors are now part of the HTML tree view — the LS sends them as part of the tree HTML payload, so they render inline in the tree rather than in the separate detail panel. No special handling needed in `BrowserHandler`.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/handlers/StopScanHandler.java
nick-y-snyk: Yes — `StopScanHandler` calls `SnykExtendedLanguageClient.getInstance().getProgressManager().cancelAll()` which cancels the in-flight LS progress tokens. The LS is notified via the work-done-progress cancellation mechanism.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/ISnykToolView.java
nick-y-snyk: Done — all 11 string constants were unused outside the interface and have been removed.

plugin/src/main/java/io/snyk/eclipse/plugin/domain/ProductConstants.java
nick-y-snyk: Done — `SCAN_PARAMS_TO_DISPLAYED` had no usages outside its own definition and has been removed. All other `ProductConstants` are still referenced.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
nick-y-snyk: Already gone — `SnykToolView` is now 28 lines after the SWT tree removal. That section no longer exists.

nick-y-snyk: Correction to my previous reply — the comment is still present, moved to `dispatchBrowserContent` (line 137) where the direct dispatch now lives. Still accurate: the HTML tree has no `ISelectionChangedListener`, so we drive the detail panel update directly.

plugin/src/main/java/io/snyk/eclipse/plugin/domain/ProductConstants.java
```diff
@@ -11,10 +11,12 @@ public final class ProductConstants {
 	public static final String SCAN_PARAMS_OSS = "oss";
 	public static final String SCAN_PARAMS_CODE = "code";
 	public static final String SCAN_PARAMS_IAC = "iac";
+	public static final String SCAN_PARAMS_SECRETS = "secrets";
 
 	public static final String DIAGNOSTIC_SOURCE_SNYK_OSS = "Snyk Open Source";
 	public static final String DIAGNOSTIC_SOURCE_SNYK_CODE = "Snyk Code";
 	public static final String DIAGNOSTIC_SOURCE_SNYK_IAC = "Snyk IaC";
+	public static final String DIAGNOSTIC_SOURCE_SNYK_SECRETS = "Snyk Secrets";
```
rrama: Super minor nitpick: Are these consts even used anymore?

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/handlers/StopScanHandler.java
rrama: Oh wow I'm so blind 🤣

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/handlers/IHandlerCommands.java
```diff
@@ -1,31 +1,11 @@
...
-	String EXPAND_ALL_TREE_NODES = "io.snyk.eclipse.plugin.commands.TreeExpand";
-	String EMPTY_CACHE = "io.snyk.eclipse.plugin.commands.snykEmptyScanResults";
-	String OPEN_PREFERENCES = "io.snyk.eclipse.plugin.properties.preferencespage";
 
-	String SHOW_CRITICAL = "io.snyk.eclipse.plugin.commands.snykFilterCritical";
-	String SHOW_HIGH = "io.snyk.eclipse.plugin.commands.snykFilterHigh";
-	String SHOW_MEDIUM = "io.snyk.eclipse.plugin.commands.snykFilterMedium";
-	String SHOW_LOW = "io.snyk.eclipse.plugin.commands.snykFilterLow";
-	String SHOW_ALL_SEVERITIES = "io.snyk.eclipse.plugin.commands.snykShowAllSeverities";
+public final class IHandlerCommands {
```
rrama: Small nit: No longer an interface, but is called `IHandlerCommands`.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
rrama: It's still here, but again, can be a follow-up...

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
rrama: Not true, but only minor nit.

tests/src/test/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClientTest.java
rrama, line 1: New stuff looks fine, I skimmed the old tests and asked AI if any gaps in coverage were introduced due to the removal of them, it said:

> Lost coverage worth noting: the deleted info-node/counts tests were the bulk of behavioral coverage for scan-result presentation. That logic moved server-side (HTML tree), so deleting them is appropriate — but there's now no client test verifying snykScan calls toolView.refreshBrowser(status) (the entire new body of snykScan). The TreeViewNotificationTest covers snykTreeView/updateTreeViewHtml, not refreshBrowser from snykScan. If you want regression protection on the new snykScan path, that's a gap.

plugin/src/main/java/io/snyk/eclipse/plugin/domain/ProductConstants.java
```diff
@@ -29,19 +31,20 @@ public final class ProductConstants {
...
-	public static final Map<String, String> SCAN_PARAMS_TO_DISPLAYED = Map.of(
-			SCAN_PARAMS_OSS, DISPLAYED_OSS,
-			SCAN_PARAMS_IAC, DISPLAYED_IAC,
-			SCAN_PARAMS_CODE, DISPLAYED_CODE_SECURITY
+	// Accepts both short codenames (navigate_to_range.go) and full product names (SnykMagnetUri/issue_enhancer.go).
+	public static final Map<String, String> PRODUCT_TO_DISPLAYED = Map.of(
+			"code", DISPLAYED_CODE_SECURITY,
+			"Snyk Code", DISPLAYED_CODE_SECURITY,
+			"secrets", DISPLAYED_CODE_SECURITY,
+			"Snyk Secrets", DISPLAYED_CODE_SECURITY,
```
rrama, line 40: Nitpick: Add a comment explaining these two, otherwise someone may think it is a bug that they map to the code display.

tests/src/test/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolViewTest.java
rrama, line 1: I asked AI about coverage gaps, it said:

> One caveat worth flagging: the selectTreeNode(...) entry point itself (tests/src/test/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolViewTest.java:27-36, 62-66) is still only covered by assertDoesNotThrow. Because selectTreeNode dispatches via Display.getDefault().asyncExec(...), which throws/swallows in headless tests, those tests don't actually exercise the dispatch — that's only reached through the dispatch* tests calling the extracted methods directly. So the replacement covers the dispatch logic but not the selectTreeNode → asyncExec → dispatch wiring end-to-end.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
rrama: I slept on it and can't think of something better that isn't overly complex, so I am going to resolve this.

rrama (approved):
Please undo your changes to `plugin/META-INF/MANIFEST.MF`. Other than that it's only minor nitpicks, so ship it!

plugin/META-INF/MANIFEST.MF
nick-y-snyk: Done — removed the duplicate org.eclipse.core.filesystem and org.eclipse.ui.ide entries; both already exist earlier in the Require-Bundle list.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Found one unused import: java.util.Set (HashSet is used but Set interface itself wasn't). Removed it.

tests/src/test/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClientTest.java
nick-y-snyk: Added two tests: snykScan_inProgress_callsRefreshBrowserWithInProgressStatus and snykScan_success_callsRefreshBrowserWithSuccessStatus — both inject toolWindowMock via setToolWindow and verify refreshBrowser is called with the correct status.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/handlers/IHandlerCommands.java
nick-y-snyk: Renamed IHandlerCommands → HandlerCommands and updated the one reference in BrowserHandler.java.

plugin/src/main/java/io/snyk/eclipse/plugin/domain/ProductConstants.java
nick-y-snyk: Added comment: // Secrets uses the Code HTML provider (shared JS runtime, no separate provider exists)

plugin/src/main/java/io/snyk/eclipse/plugin/domain/ProductConstants.java
nick-y-snyk: Removed all four DIAGNOSTIC_SOURCE_SNYK_* constants — none were used in prod code. The one test reference (SnykUriDetailsTest) compared against DIAGNOSTIC_SOURCE_SNYK_CODE; replaced with the literal "Snyk Code" and dropped the import.

PR #418, opened 2026-06-10
fix(restart): restart Eclipse on Snyk CLI path change [IDE-824]

acke:
## Summary

Stacked on top of [fix/IDE-1690-windows-arm-install](https://github.com/snyk/snyk-eclipse-plugin/tree/fix/IDE-1690-windows-arm-install). When the user changes the Snyk CLI binary path in the HTML settings page, the running language server is still bound to the old binary. In-process LS restart via lsp4e proved unreliable (cached \`StreamConnectionProvider\`, stuck failed wrappers, protocol-version probes against the wrong CLI), so this PR takes the simpler route: **prompt the user to restart Eclipse on CLI path change, and trigger the restart on confirmation**.

- \`HTMLSettingsPreferencePage.handleSave\` captures the previous CLI path before applying settings and, if it changed, calls \`SnykLanguageServer.promptToRestartEclipseForNewCli()\`.
- \`promptToRestartEclipseForNewCli\` now opens a real \`MessageDialog.openQuestion\` on the UI thread; on Yes it calls \`PlatformUI.getWorkbench().restart()\`. The previous implementation only posted a transient \`logAndShow\` INFO status with no actionable prompt and no restart trigger.

## Test plan

- [x] Built update site and installed in Eclipse on macOS.
- [x] Changed CLI Path in Snyk preferences -> prompt appears -> choosing Yes restarts Eclipse -> new CLI is in effect on next launch.
- [ ] CI green on PR.


rrama (approved):
Might want to add a test. Here is what AI said:

> What's easy to test
> The CLI-path-change detection in parseAndSaveConfig (HTMLSettingsPreferencePage.java:233-241) fits the existing test pattern cleanly. There's already a test class HTMLSettingsPreferencePageTest.java that uses an in-memory Preferences store and invokes the private parseAndSaveConfig via reflection (e.g. parseAndSaveConfig_savesCliPath).
> 
> You'd add a test that:
> 
> Seeds an initial CLI path in the in-memory store.
> Calls parseAndSaveConfig with JSON containing a different cli_path.
> Verifies SnykLanguageServer.promptToRestartEclipseForNewCli() was/wasn't called.
> To assert the call, you'd wrap the static method in a Mockito mockStatic(SnykLanguageServer.class) — the repo already does this pattern (see SnykLoggerTest.java:17 mocking StatusManager). So a "changed → prompts" and "unchanged → does not prompt" pair of tests is straightforward.
> 
> What's hard / can't really be unit-tested
> The body of promptToRestartEclipseForNewCli (SnykLanguageServer.java:141-153) is almost pure Eclipse UI plumbing:
> 
> PlatformUI.getWorkbench().getDisplay()
> display.asyncExec(...)
> MessageDialog.openQuestion(...)
> PlatformUI.getWorkbench().restart()
> In a headless unit test there's no running Workbench, and the work is deferred to asyncExec with no display loop. You could mockStatic(PlatformUI.class) and mockStatic(MessageDialog.class), but you'd be asserting against a mock you fully control, with the asyncExec lambda never naturally executing — low-value and brittle. The real "Yes → restart" behavior is only verifiable manually (which the PR's test plan does).
> 
> Practical recommendation
> The method as written mixes the decision (path changed) with un-testable UI side effects. The decision lives in the preference page and is testable. So the realistic, valuable test is the one in HTMLSettingsPreferencePageTest verifying the prompt is triggered only when the path actually changes — covering the core regression risk of this PR with minimal effort.
> 
> One note: the PR description refers to a method handleSave, but the change actually lives in parseAndSaveConfig (HTMLSettingsPreferencePage.java:192-245). No handleSave method exists.

PR #419, opened 2026-06-10
fix(download): unblock Windows-on-ARM install [IDE-1690]

acke:
> Replaces #417, which was accidentally merged into `fix/IDE-824-restart-ls-on-cli-change` instead of `main`. Same branch, same commits — re-opened against the correct base so the IDE-1690 changes can actually be reviewed and shipped.

## Summary

Two follow-on fixes after testing the IDE-1690 ARM filename fallback on a real Windows-on-ARM machine — the filename fix alone wasn't enough to get the CLI downloaded on a fresh install.

- **`25d2025`** — drop the `-arm64` suffix on Windows so the filename matches the published `snyk-win.exe` asset (no `snyk-win-arm64.exe` exists; Windows-11-on-ARM runs the x64 binary via emulation).
- **`87874d9`** — coerce empty/blank `CLI_PATH` and `CLI_BASE_URL` prefs back to their defaults. Both prefs are seeded with defaults via `PreferenceStore.setDefault`, but those only apply when the key has *no* stored value. When stored as an empty string the empty value wins, which produced:
  - `CLI_PATH=""` → `new File("")` → `getParentFile()` is `null` → NPE during `mkdirs()` in `SnykStartup.download`.
  - `CLI_BASE_URL=""` → `String.format("%s/cli/…", "", …)` → relative URI `/cli/<v>/<asset>` → Apache HttpClient throws `Target host is not specified`, surfaced as a generic `ClientProtocolException`.
- Same commit also walks the cause chain in `logDownloadFailure`, so the real transport error (SSLException / ProtocolException / etc.) shows up in the error log instead of just the wrapper class name. `getVersion`/`getSha` now include the URL in the thrown exception message.

Neither pref problem is arch-specific — they reproduce on any platform with blank prefs.

## Test plan

- [x] Unit tests: `LsRuntimeEnvironmentTest` covers win-arm → x64 fallback and mac-arm pass-through.
- [x] Manual verification on a Windows-on-ARM Eclipse: the binary now downloads on first launch and the preferences page shows the resolved CLI path.
- [ ] CI green on PR.
- [ ] Quick smoke on Windows x64 / macOS / Linux to confirm no regression in the default-pref path.


plugin/src/main/java/io/snyk/languageserver/download/LsBinaries.java
```diff
@@ -9,12 +9,24 @@
 public class LsBinaries {
   private static final Preferences PREFERENCES = Preferences.getInstance();
   public static final String REQUIRED_LS_PROTOCOL_VERSION = "25";
+  // Canonical CDN host. Used when the CLI_BASE_URL preference is unset or blank
+  // — otherwise the formatted URL has no host and HttpClient throws
+  // "Target host is not specified".
+  public static final String DEFAULT_CLI_BASE_URL = "https://downloads.snyk.io";
```
rrama: This is not the only place we hardcode "https://downloads.snyk.io". Please make them all use the same const.

plugin/src/main/java/io/snyk/eclipse/plugin/SnykStartup.java
```diff
@@ -1,180 +1 @@
...
-
-	public static boolean isDownloading() {
-		return downloading;
-	}
-
-	public void setLogger(ILog logger) {
-		SnykStartup.logger = logger;
-	}
-}
+@-
```
rrama: Umm... Finger slip? xP

plugin/src/main/java/io/snyk/eclipse/plugin/SnykStartup.java
```diff
@@ -91,20 +91,25 @@ private void downloadIfNeeded(IProgressMonitor monitor) {
...
+				logDownloadFailure(errorMessage.toString());
 			}
 
 			private void logDownloadFailure(String errorMessage) {
 				// Log the error - user may just be offline but have an existing binary. We show a user-facing error later if binary is missing.
 				String message = "Failed to download Snyk CLI: " + errorMessage
 						+ ". Will try to start with existing binary if available.";
 				logger.error(message);
-				SnykLogger.logAndShowWarning("Snyk CLI download failed: " + errorMessage
-						+ ". Will try to start with existing binary if available.");
```
rrama, line 107: Have you manually confirmed that the other error that is shown is nicer than this one? Is this the one that links to the error log or is that the other one?

plugin/src/main/java/io/snyk/eclipse/plugin/SnykStartup.java
acke: Not a slip — the file was accidentally corrupted to a single `@-` byte on the branch (likely a bad conflict resolution during an earlier rebase). Commit `711cd905` restored the full file content and applied the two PMD fixes (reference equality and null assignment in the cause-chain loop) at the same time. The net change vs `main` is only the `logDownloadFailure` cause-chain walk and the CLI_PATH persistence from `8f5b524a` — nothing unintended.

plugin/src/main/java/io/snyk/languageserver/download/LsBinaries.java
acke: Fixed — `95f12d3b` (refactor: centralize CLI base URL constant) removed the last hardcoded `"https://downloads.snyk.io"` from `LsBinaries` and made it reference `Preferences.DEFAULT_CLI_BASE_URL`. `19d6e711` then removed the now-duplicate local constant that `LsBinaries` had defined itself. All callers now go through the single constant in `Preferences`.

plugin/src/main/java/io/snyk/eclipse/plugin/SnykStartup.java
```diff
@@ -158,12 +163,19 @@ static LsDownloader getLsDownloader() throws URISyntaxException {
...
-		final File lsFile = new File(Preferences.getInstance().getCliPath());
+		final String cliPath = Preferences.getInstance().getCliPath();
+		final File lsFile = new File(cliPath);
 		try {
 			LsDownloader lsDownloader = getLsDownloader();
 			lsFile.getParentFile().mkdirs();
 			lsDownloader.download(monitor);
 			lsFile.setExecutable(true);
+			// Persist the effective path so it shows up in the preference page when the
+			// stored value was empty (and we fell back to the default location).
```
rrama: Super nitpicky: You already fixed the preference page above, so the comment is a slight lie. But it is good that we are persisting on a successful download. Although I am now thinking we should persist not on a successful download, but a successful find of the binary at that location? So it includes people who disable automatic downloads. I doubt we'd change the default in the future without migration code, so I am probably overthinking this all.

tests/src/test/java/io/snyk/languageserver/LsRuntimeEnvironmentTest.java
rrama, line 53: This test needs updating too.

rrama (approved):
LGTM!

PR #420, opened 2026-06-10
fix: notify LS when workspace projects are added or removed [IDE-910]

acke:
## Summary
- Plugin never registered an `IResourceChangeListener`, so importing/closing/deleting projects didn't propagate to the language server. LS kept scanning the workspace folder set received at initialization → stale issues for removed projects, no results for newly imported ones until Eclipse restart.
- Adds `WorkspaceFolderChangeTracker` listening on `POST_CHANGE`, `PRE_CLOSE`, `PRE_DELETE`; recomputes top-level project set and emits `workspace/didChangeWorkspaceFolders` LSP notification with the diff.
- Registered from `SnykStartup` right after LS startup.

## Test plan
- [ ] Open Eclipse with no workspace projects → verify LS scans nothing
- [ ] Import a project → verify LS receives `didChangeWorkspaceFolders` (added) and scans it
- [ ] Close a project → verify LS receives notification (removed) and clears its issues
- [ ] Delete a project → verify LS receives notification (removed) and clears its issues
- [ ] Re-open a closed project → verify LS receives notification (added) and rescans


plugin/src/main/java/io/snyk/eclipse/plugin/SnykStartup.java
```diff
@@ -56,6 +57,7 @@ protected IStatus run(IProgressMonitor monitor) {
 
 				try {
 					SnykLanguageServer.startSnykLanguageServer();
+					WorkspaceFolderChangeTracker.register();
```
bastiandoetsch, line 60: [Critical] `register()` is called here but `WorkspaceFolderChangeTracker.unregister()` is never called on plugin shutdown or reload.

If the plugin is disabled and re-enabled in the same Eclipse session, `register()` returns early at the `instance != null` guard — but the original listener is still registered with `ResourcesPlugin`, so two listeners will fire on every workspace change, sending duplicate notifications to the language server. The `unregister()` method exists at `WorkspaceFolderChangeTracker.java:52` but has no caller.

Fix: wire `WorkspaceFolderChangeTracker.unregister()` to the plugin's shutdown lifecycle — either via `BundleActivator.stop()` or by registering a bundle listener in `SnykStartup`.

— AI review

plugin/src/main/java/io/snyk/languageserver/WorkspaceFolderChangeTracker.java
```diff
@@ -0,0 +1,127 @@
...
+            for (Map.Entry<String, WorkspaceFolder> e : knownFolders.entrySet()) {
+                if (!currentFolders.containsKey(e.getKey())) removed.add(e.getValue());
+            }
+            if (added.isEmpty() && removed.isEmpty()) {
+                return;
+            }
+            knownFolders.clear();
+            knownFolders.putAll(currentFolders);
+        }
+        sendDidChange(added, removed);
```
bastiandoetsch: [Critical] `knownFolders` is updated at lines 86–87 inside the lock, but `sendDidChange` is called outside the lock at line 89 and silently returns when the LS client is `null` (e.g. during the async startup window between `startSnykLanguageServer()` returning and the LS fully initializing). When that happens, the tracker's internal state already reflects the new folder but the LS was never told — so the next folder change will compute a delta against the updated state, permanently missing the notification for the change that was silently dropped.

Concrete scenario: user imports a project in the seconds between plugin startup and LS initialization completing. The tracker fires, updates `knownFolders`, `client == null` → silent return. LS initializes, unaware of the new project.

Fix: either (a) defer `register()` until the LS client is connected (subscribe to LS lifecycle events), or (b) don't update `knownFolders` when the notification cannot be sent (roll back the state update), or (c) re-sync folder state when the LS comes online.

— AI review

plugin/src/main/java/io/snyk/languageserver/WorkspaceFolderChangeTracker.java
```diff
@@ -0,0 +1,127 @@
...
+    private final Object lock = new Object();
+
+    private WorkspaceFolderChangeTracker() {
+        snapshot();
+    }
+
+    public static synchronized void register() {
+        if (instance != null) return;
+        instance = new WorkspaceFolderChangeTracker();
+        ResourcesPlugin.getWorkspace().addResourceChangeListener(instance, CHANGE_MASK);
```
bastiandoetsch: [Should Fix] `instance` is assigned at line 48 before `addResourceChangeListener` is called at line 49. If `addResourceChangeListener` throws (e.g. workspace not yet fully started), `instance` is non-null but the listener was never registered. A subsequent `register()` call returns early at line 47 (`instance != null`), silently leaving the plugin in a state where it believes the listener is active but it isn't.

Fix: assign `instance` after the listener is successfully registered:
```java
public static synchronized void register() {
    if (instance != null) return;
    WorkspaceFolderChangeTracker tracker = new WorkspaceFolderChangeTracker();
    ResourcesPlugin.getWorkspace().addResourceChangeListener(tracker, CHANGE_MASK);
    instance = tracker; // only set after successful registration
}
```

— AI review

plugin/src/main/java/io/snyk/eclipse/plugin/SnykStartup.java
acke: Fixed in de4bc3e7 — `WorkspaceFolderChangeTracker.unregister()` is now called from `Activator.stop()`. The bundle activator's `stop()` is the canonical shutdown hook for OSGi plugins and fires on both Eclipse shutdown and plugin disable/re-enable in the same session, so duplicate listener registration is no longer possible.

plugin/src/main/java/io/snyk/languageserver/WorkspaceFolderChangeTracker.java
acke: Fixed in de4bc3e7 — `knownFolders` is no longer updated before we know the notification was sent. `sendDidChange` now returns `boolean` (true = notification dispatched, false = client/LS unavailable or transport error). `recomputeAndNotify` only calls `knownFolders.clear()/putAll` in a second synchronized block after `sendDidChange` returns true, so a project imported before the LS is connected stays in the pending delta and will be included in the next notification once the LS is up.

plugin/src/main/java/io/snyk/languageserver/WorkspaceFolderChangeTracker.java
acke: Fixed in de4bc3e7 — `register()` now constructs the tracker and calls `addResourceChangeListener` before assigning `instance`. If the listener registration throws, `instance` stays null and the next `register()` call retries correctly.

plugin/src/main/java/io/snyk/languageserver/WorkspaceFolderChangeTracker.java
```diff
@@ -0,0 +1,134 @@
...
+ * when the set of top-level accessible projects changes. Without this, the LS
+ * keeps scanning the workspace folders it was told about at initialization
+ * time, so newly imported projects are ignored and removed projects keep
+ * producing stale results.
+ */
+@SuppressWarnings("restriction")
+public final class WorkspaceFolderChangeTracker implements IResourceChangeListener {
+
+    private static final int CHANGE_MASK =
+            IResourceChangeEvent.POST_CHANGE | IResourceChangeEvent.PRE_CLOSE | IResourceChangeEvent.PRE_DELETE;
```
rrama: Is `POST_CHANGE` too expensive to not have further filtering on? And does it not cover for close and delete as well?

AI has suggested:
```java
public void resourceChanged(IResourceChangeEvent event) {
    if (event.getType() != IResourceChangeEvent.POST_CHANGE) return;

    IResourceDelta rootDelta = event.getDelta();
    if (rootDelta == null) return;

    // Only look at direct children of the Workspace Root (which are Projects)
    IResourceDelta[] projectDeltas = rootDelta.getAffectedChildren();
    
    for (IResourceDelta delta : projectDeltas) {
        IResource resource = delta.getResource();
        
        // Ensure we are strictly dealing with a Project
        if (resource instanceof IProject) {
            int kind = delta.getKind();
            int flags = delta.getFlags();

            // 1. Project was completely added/created
            if (kind == IResourceDelta.ADDED) {
                System.out.println("Project Added: " + resource.getName());
            }
            
            // 2. Project was completely deleted
            else if (kind == IResourceDelta.REMOVED) {
                System.out.println("Project Removed: " + resource.getName());
            }
            
            // 3. Project was Opened or Closed (State Changed)
            else if (kind == IResourceDelta.CHANGED && (flags & IResourceDelta.OPEN) != 0) {
                IProject project = (IProject) resource;
                if (project.isOpen()) {
                    System.out.println("Project Opened: " + project.getName());
                } else {
                    System.out.println("Project Closed: " + project.getName());
                }
            }
        }
    }
}
```


| Scenario | Event Type | Target Delta Logic |
| :--- | :--- | :--- |
| **Project Created** | `POST_CHANGE` | `delta.getKind() == IResourceDelta.ADDED` |
| **Project Opened** | `POST_CHANGE` | `(delta.getFlags() & IResourceDelta.OPEN) != 0 && project.isOpen()` |
| **Project Closed** | `PRE_CLOSE` or `POST_CHANGE` | `(delta.getFlags() & IResourceDelta.OPEN) != 0 && !project.isOpen()` |
| **Project Deleted** | `PRE_DELETE` or `POST_CHANGE` | `delta.getKind() == IResourceDelta.REMOVED` |

acke: Good point on both counts. Fixed in 2f9d324:

- Narrowed the mask to `POST_CHANGE` only — `PRE_CLOSE` and `PRE_DELETE` are redundant since `POST_CHANGE` covers those via `delta.getKind() == REMOVED` and `(delta.getFlags() & IResourceDelta.OPEN) != 0`.
- Added `affectsProjectSet()` which checks only the direct children of the workspace root delta (always top-level projects) and returns false unless a project is ADDED, REMOVED, or has the OPEN flag set. File saves, compilations, marker changes, etc. all return false and skip `recomputeAndNotify()` entirely.

acke:
All CI checks pass and review discussion resolved. Blocked on required code owner approval from `snyk/developer-experience_ide` before merge.

plugin/src/main/java/io/snyk/languageserver/WorkspaceFolderChangeTracker.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    // POST_CHANGE fires for every resource event (file saves, compilations, etc).
+    // Only proceed when top-level projects are added, removed, or opened/closed.
+    private static boolean affectsProjectSet(IResourceDelta rootDelta) {
+        if (rootDelta == null) return false;
+        for (IResourceDelta delta : rootDelta.getAffectedChildren()) {
+            int kind = delta.getKind();
+            if (kind == IResourceDelta.ADDED || kind == IResourceDelta.REMOVED) return true;
+            if ((delta.getFlags() & IResourceDelta.OPEN) != 0) return true;
```
rrama, line 93: Please add a comment explaining that a close of a project also adds the `OPEN` flag, this was not apparent to my non-AI mind.

rrama (commented):
Code LGTM, other than needing a comment explaining `IResourceDelta.OPEN`. Please just add some tests to prove the core logic.

tests/src/test/java/io/snyk/languageserver/WorkspaceFolderChangeTrackerTest.java
```diff
@@ -0,0 +1,86 @@
...
+            rootDelta(child(IResourceDelta.CHANGED, IResourceDelta.OPEN))));
+    }
+
+    @Test
+    void closeFlagIsOpenFlag_returnsTrue() {
+        // Closing a project sets IResourceDelta.OPEN — it is the same bit used
+        // for open. This test makes the counterintuitive mapping explicit.
+        assertTrue(WorkspaceFolderChangeTracker.affectsProjectSet(
+            rootDelta(child(IResourceDelta.CHANGED, IResourceDelta.OPEN))));
+    }
```
rrama: The comment in the test above was fine, I don't need the same test twice.

rrama (commented):
Still missing unit tests for the logic I actually cared about `recomputeAndNotify`. And some more integration-esque tests of adding and removing and ensuring the notifications are sent and the transitions caused the correct states.

tests/src/test/java/io/snyk/languageserver/WorkspaceFolderChangeTrackerTest.java
```diff
@@ -83,4 +111,196 @@ void mixedChildren_oneAdded_returnsTrue() {
...
+        AtomicReference<Map<String, WorkspaceFolder>> source =
+            new AtomicReference<>(Map.of("file:///a", a));
+        List<List<WorkspaceFolder>> sentAdded = new ArrayList<>();
+        List<List<WorkspaceFolder>> sentRemoved = new ArrayList<>();
+
+        WorkspaceFolderChangeTracker tracker = trackerWith(source, sentAdded, sentRemoved, true);
+
+        tracker.recomputeAndNotify();
+
+        assertTrue(sentAdded.isEmpty(), "no send when nothing changed");
```
rrama, line 131: `assertTrue(sentRemoved.isEmpty(), "no send when nothing changed");` as well?
Same for below.

rrama (approved):

PR #422, opened 2026-06-12
fix(version-check): block LS launch on incompatible CLI binary with clear error [IDE-2112]

acke:
## Summary

- Blocks Language Server launch when CLI binary is incompatible: `startSnykLanguageServer()` now runs `verifyCliProtocolVersion()` **before** calling lsp4e, so an incompatible CLI no longer triggers the generic Eclipse "Problem Occurred" dialog
- Shows a clean **"Snyk CLI version incompatible"** dialog (SWT `MessageDialog` with a `Link` widget) listing required vs actual protocol version and linking to the [IDE Plugin Compatibility Matrix](https://docs.snyk.io/developer-tools/snyk-ide-plugins-and-extensions/compatibility-matrix)
- `startSnykLanguageServer()` returns `boolean`; `SnykStartup` gates the Snyk Wizard on `true` — no wizard if CLI is incompatible
- `ensureLanguageServerRunning()` retry loop breaks immediately on `false` return — prevents repeated dialog popups every second
- Raw CLI output and NFE cause chain kept out of user-facing messages (debug-only logging)
- Parses only the first output line so binaries that print version + trailing warnings are handled correctly

Closes IDE-2112

## Test plan
- [ ] Install from update site with an incompatible CLI binary → single "Snyk CLI version incompatible" dialog with compatibility-matrix link; no "Problem Occurred" dialog; no Snyk Wizard
- [ ] Install with CLI path pointing to non-existent binary → "binary not found" error; no wizard
- [ ] Install with a compatible CLI → normal startup + wizard (if unauthenticated)
- [ ] Clicking the link in the incompatible-CLI dialog opens the browser
- [ ] CI green


.cursor/skills/ide-pr-testing-strategy/SKILL.md
```diff
@@ -0,0 +1,177 @@
+---
+name: ide-pr-testing-strategy
```
acke: Remove

plugin/src/main/java/io/snyk/eclipse/plugin/utils/SnykLogger.java
```diff
@@ -15,6 +15,12 @@ public static void logError(Exception exception) {
 		StatusManager.getManager().handle(status, StatusManager.SHOW | StatusManager.LOG);
 	}
 
+	public static void logDebug(String message) {
+		Status status = new Status(IStatus.INFO, Activator.PLUGIN_ID, message);
+
+		StatusManager.getManager().handle(status, StatusManager.LOG);
+	}
```
rrama: This is the same function as `logInfo`...

.cursor/skills/ide-pr-testing-strategy/SKILL.md
rrama, line 1: Is this meant to be added? I don't want to review it if not.

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
```diff
@@ -55,22 +61,32 @@ public void start() throws IOException {
...
+		String output = "";
 		try {
 			ProcessBuilder pb = new ProcessBuilder(cliPath, "language-server", "--protocolVersion");
 			pb.redirectErrorStream(true);
 			Process proc = pb.start();
-			String output = new String(proc.getInputStream().readAllBytes(), StandardCharsets.UTF_8).trim();
+			output = new String(proc.getInputStream().readAllBytes(), StandardCharsets.UTF_8).trim();
 			proc.waitFor(5, TimeUnit.SECONDS);
-			int actual = Integer.parseInt(output);
+			// Use first line only: some CLI builds print version + trailing warnings.
```
rrama: Can you give an example? I see in the tests below it can say "Warning: something deprecated", if so just add that to the comment.

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
```diff
@@ -55,22 +61,32 @@ public void start() throws IOException {
...
-			SnykLogger.logError(e);
+		} catch (NumberFormatException e) {
+			int expected = Integer.parseInt(io.snyk.languageserver.download.LsBinaries.REQUIRED_LS_PROTOCOL_VERSION);
+			showIncompatibleCliDialog(expected, -1);
+			SnykLogger.logDebug("verifyCliProtocolVersion: unexpected output: " + output + " (" + e.getMessage() + ")");
+			String msg = "Snyk CLI binary at '" + cliPath + "' is not compatible with this version of the Eclipse plugin. Please update the Snyk CLI.";
+			throw new IOException(msg); //NOPMD - PreserveStackTrace: NFE cause intentionally dropped to keep raw CLI output out of lsp4e's "Problem Occurred" dialog (IDE-2112)
+		} catch (InterruptedException e) {
+			Thread.currentThread().interrupt();
+			SnykLogger.logInfo("CLI protocol version check interrupted.");
```
rrama: Then don't allow it on like everything is hunky dorey. This is a bad path and we need to not run the LS!

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
```diff
@@ -121,10 +137,49 @@ private static void waitForDownload() {
 		}
 	}
 
-	public static void startSnykLanguageServer() {
+	public static boolean startSnykLanguageServer() {
+		Preferences prefs = Preferences.getInstance();
+		try {
+			String cliPath = getCliPathOrThrow(prefs);
+			verifyCliProtocolVersion(cliPath);
+		} catch (IOException e) {
```
rrama, line 150: I do not like using a catch to verify things, especially on a generic exception. Can we just just return a boolean up the stack?

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
```diff
@@ -40,7 +47,6 @@ public void start() throws IOException {
 		waitForInit();
 
 		String cliPath = getCliPathOrThrow(prefs);
-		verifyCliProtocolVersion(cliPath);
```
rrama, line 43: I know we don't have automatic restarts today, but if we do add automatic restarts, then this change could become a gap...

rrama (commented):
I am not comfortable with the "Didn't throw therefore the binary is safe" technique, I would much rather we just return a boolean (or even a two state enum) up the stack, that way we are 100% sure we haven't hit some shortcut that missed the check.

Also, please attach screenshots of the new error message.

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
acke: Fixed — `InterruptedException` now throws `IOException` so the LS does not start when the version check is interrupted.

acke:
Refactored `verifyCliProtocolVersion` to return `boolean` (true = compatible, false = not) instead of throwing for compatibility failures. `startSnykLanguageServer` now checks the boolean directly. Tests updated. For screenshots — could you add those yourself to the PR description?

plugin/src/main/java/io/snyk/eclipse/plugin/utils/SnykLogger.java
acke: Removed `logDebug` — it was identical to `logInfo`. Call site updated to use `logInfo`.

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
acke: Added `verifyCliProtocolVersion` back to `start()` — restarts now also check the version. Since `verifyCliProtocolVersion` returns `boolean` and shows the incompatible dialog itself, the `IOException` thrown to lsp4e contains only a clean message (no raw CLI output).

.cursor/skills/ide-pr-testing-strategy/SKILL.md
acke: Removed — accidentally committed, shouldn't be in this PR.

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
acke: Added example: `"Warning: something deprecated"` — matches the test mock.

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
acke: Addressed — `verifyCliProtocolVersion` now returns `boolean` instead of throwing for compatibility failures. The remaining `catch(IOException)` in `startSnykLanguageServer` covers only `getCliPathOrThrow` which throws for a genuinely exceptional case (CLI binary not found on disk).

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
```diff
@@ -55,22 +55,33 @@
...
 			}
-		} catch (NumberFormatException | InterruptedException e) {
-			SnykLogger.logError(e);
+		} catch (NumberFormatException e) {
+			String truncated = output.length() > 80 ? output.substring(0, 80) + "..." : output;
+			String msg = "Snyk CLI binary at '" + cliPath + "' is not compatible: "
+					+ "'snyk language-server --protocolVersion' did not return a version number "
+					+ "(got: '" + truncated + "'). Please update the Snyk CLI.";
+			SnykLogger.logAndShowError(msg);
+			throw new IOException(msg);
```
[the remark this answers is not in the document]
acke: No longer applicable — the `NumberFormatException` catch now returns `false` instead of throwing, so there's no exception whose stack trace needs preserving.

plugin/src/main/java/io/snyk/languageserver/SnykLanguageServer.java
acke: Update: after the boolean refactor, `InterruptedException` now returns `false` (rather than throwing) — both callers (`start()` and `startSnykLanguageServer()`) treat `false` as "do not start the LS". Log message: `"CLI protocol version check interrupted — not starting Language Server."`

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
```diff
@@ -123,6 +124,7 @@ protected void outputCommandResult(Object result) {
 	@Override
 	public void selectTreeNode(String issueId, String product) {
 		if (issueId == null || issueId.isEmpty()) return;
+		if (Preferences.getInstance().isTest()) return;
```
rrama, line 127: Nitpick: TBH, I would have prefered DI or mocking or test classes extending the real ones which override the throwing parts.
But this will do.

<details><summary>AI's take:</summary>

1. isTest() branch in production code (current approach)
Pros

Trivial to write; zero test scaffolding.
Works even where construction is hard to intercept (a ViewPart instantiated by the Eclipse framework).
Cons

Test logic ships in production. Every call pays an isTest() lookup and there's a live code path that only exists for tests.
Reduced coverage of the real branch. The test exercises the early-return, not dispatchTreeNode/drain logic. Note the tests that do verify real behavior (dispatchTreeNode_*) call the lower-level methods directly and inject mocks via reflection (SnykToolViewTest.java:70-126) — they bypass the guard entirely. So the guard mainly protects the two assertDoesNotThrow smoke tests (:28-37), which the existing try/catch arguably already covers.
Relies on global mutable state (Preferences.getInstance().setTest(true)), which can leak across tests if setup/teardown is imperfect.
2. Test subclass overriding the "dangerous" methods
e.g. override a protected runOnUiThread(Runnable) / getDisplay() seam to be synchronous/no-op.

Pros

Production code stays clean — no isTest() conditionals.
Forces you to extract a UI seam (e.g. an injectable "UI executor"), which improves the design.
Real logic inside the runnable can still be exercised synchronously.
Cons

Requires refactoring to expose overridable seams; Display.getDefault() is a static call today, so you'd need to introduce that indirection first.
Subclass-override testing is brittle (tests coupled to method structure).
3. Mocking / injecting a UI abstraction (DI)
Introduce something like an UiSynchronizer interface (asyncExec(Runnable)), inject the real one in production and a synchronous mock in tests.

Pros

Cleanest separation; production has no test awareness.
Lets tests assert that the runnable did the right thing synchronously (real coverage), and verify interactions.
Removes the static Display.getDefault() dependency entirely.
Cons

Biggest upfront refactor; SnykToolView is framework-instantiated so you'd need a settable/default-injected collaborator.
Mocking SWT widgets (Browser, TreeViewBrowserHandler) is still awkward — but those are already handled here via reflection injection.
4. Headless SWT Display in tests (SWTBot-style)
Run with a real off-screen Display.

Pros: exercises the genuine path. Cons: heavy, platform-native dependencies, flaky in CI, slow — usually not worth it for unit tests.

My take
Given the try/catch already neutralizes the headless-crash risk, the isTest() guards add little and import test-awareness into production for marginal benefit. The cleanest of the options is #3 (inject a UI synchronizer), which would let the tests drop both the isTest() guards and the broad catch blocks while actually covering the real branch. If a full refactor isn't desired, the pragmatic minimum is to keep the existing try/catch and drop the isTest() checks — or at least acknowledge they're redundant rather than "necessary."

One caveat: I'm inferring the headless-crash behavior from the catch block's intent and the test structure; I haven't run the suite. If there's a path where class-loading of Display itself fails before entering the try, the guard would matter — but in both methods the Display.getDefault() call is inside the try, so that doesn't apply here.
</details>

rrama (approved):

PR #424, opened 2026-06-15
fix(IDE-2112): show clean incompatible-CLI dialog, suppress Problem Occurred and wizard

acke:
## Summary

- Adds `verifyCliProtocolVersion()` check **before** invoking lsp4e, so an incompatible CLI no longer triggers the generic Eclipse "Problem Occurred" dialog
- `startSnykLanguageServer()` now returns `boolean`; callers gate downstream behaviour on it
- Snyk Wizard is suppressed when the CLI check fails — no point authenticating against a non-functional LS
- `ensureLanguageServerRunning()` retry loop breaks immediately on `false` return, preventing repeated incompatibility dialogs every second
- Dialog message uses plain language ("requires protocol version X, but installed CLI reports version Y") instead of leaking variable names
- Raw CLI output and NFE cause chain kept out of user-facing messages (debug-only logging)

## Test plan

- [ ] Install plugin from update site built on this branch (`file:///Users/skippy/repos/snyk-eclipse-plugin2/update-site/target/repository`)
- [ ] Point CLI path to an incompatible binary (wrong protocol version) → verify single "Snyk CLI version incompatible" dialog with compatibility-matrix link, no "Problem Occurred" dialog, no Snyk Wizard
- [ ] Point CLI path to a non-existent path → verify "binary not found" error, no wizard
- [ ] Point CLI path to a valid compatible CLI → verify normal startup + wizard (if unauthenticated)
- [ ] Verify clicking the compatibility-matrix link in the dialog opens the browser

### DIFF
--- a/.cursor/skills/ide-pr-testing-strategy/SKILL.md
+++ b/.cursor/skills/ide-pr-testing-strategy/SKILL.md
@@ -1,0 +1,177 @@
+---
+name: ide-pr-testing-strategy
+description: >-
+  Reviews IDE plugin PRs against the Team IDE Testing Strategy (LDX-Sync / config
+  sync) and suggests automated tests, CI gates, and manual QA. Use when reviewing
+  a pull request, asking for test coverage improvements, QA plan for IDE-1652, or
+  mapping changes to the Confluence Testing Strategy page.
+---
+
+# IDE PR Testing Strategy Review
+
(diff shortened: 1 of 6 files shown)


acke:
Duplicate — correct PR is #423 (stacked on #422). Closed in favour of that one.

PR #426, opened 2026-06-17
feat: per-folder reset overrides [IDE-1945]

nick-y-snyk:
### Description

IDE-side support for the per-folder **Reset overrides** button (the button + dialog are served by snyk-ls). See snyk/snyk-ls#1344 for the LS side and the shared `FOLDER_RESET_FIELDS` list this mirrors.

A folder field sent by the dialog as JSON `null` means "reset this override" → the LS `Unset`s the `user:folder:` override (fallback to org/LDX/default). Native nullable types cannot distinguish present-`null` from absent, so the plugin re-parses the raw JSON tree and, for the reset-key whitelist, forwards `{value:null, changed:true}`.

This PR extends that whitelist to the 3 fields snyk-ls recently added so reset clears them too: `additional_parameters`, `additional_environment`, `scan_command_config`. See the commit body for plugin-specific details (naming fixes, stored-override clearing).

### Checklist

- [ ] README updated, if user-facing


tests/src/test/java/io/snyk/eclipse/plugin/preferences/HTMLSettingsPreferencePageTest.java
```diff
@@ -296,6 +305,144 @@ void parseAndSaveConfig_mixedNullAndValueFields() throws Exception {
...
+		assertEquals(Boolean.TRUE, setting.getChanged());
+	}
+
+	@Test
+	void parseAndSaveConfig_anyFolderFieldNullIsForwardedAsReset() throws Exception {
+		FolderConfigSettings.setInstance(new FolderConfigSettings());
+		String folderPath = "/work/basebranch-project";
+		// The IDE no longer maintains a reset whitelist: every folder field sent as JSON null is
+		// forwarded as {value:null, changed:true}. snyk-ls is authoritative and ignores nulls on
+		// keys with no fallback layer, so forwarding e.g. base_branch is a safe no-op there.
```
rrama: AI comment giving me a history lesson...

tests/src/test/java/io/snyk/eclipse/plugin/preferences/HTMLSettingsPreferencePageTest.java
```diff
@@ -296,6 +305,144 @@ void parseAndSaveConfig_mixedNullAndValueFields() throws Exception {
...
+		LspFolderConfig stored = FolderConfigSettings.getInstance().getFolderConfig(folderPath);
+		assertResetSetting(stored, "base_branch");
+	}
+
+	@Test
+	void parseAndSaveConfig_folderNonNullArrayAndObjectAreStored() throws Exception {
+		FolderConfigSettings.setInstance(new FolderConfigSettings());
+		String folderPath = "/work/stored-project";
+		// Non-null array (:301-306): stored as List<String>, with the null element filtered out (:303).
+		// Non-null scan_command_config object (:288-300): deserialized to Map<String, ScanCommandConfig>.
```
rrama: AI comment referring to line numbers directly.

tests/src/test/java/io/snyk/eclipse/plugin/preferences/HTMLSettingsPreferencePageTest.java
```diff
@@ -296,6 +305,144 @@ void parseAndSaveConfig_mixedNullAndValueFields() throws Exception {
...
+		assertEquals("echo pre", oss.preScanCommand());
+		assertTrue(oss.preScanOnlyReferenceFolder());
+		assertEquals("echo post", oss.postScanCommand());
+		assertFalse(oss.postScanOnlyReferenceFolder());
+	}
+
+	@Test
+	void parseAndSaveConfig_folderConfigWithoutPathIsIgnored() throws Exception {
+		FolderConfigSettings.setInstance(new FolderConfigSettings());
+		// folderPath absent -> processFolderConfig returns early (:266-269): nothing stored, no throw.
```
rrama: AI comment referring to line numbers directly.

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/HTMLSettingsPreferencePage.java
```diff
@@ -273,7 +302,16 @@
         var field = fields.next();
         String key = field.getKey();
         JsonNode node = field.getValue();
-        if ("folderPath".equals(key) || node.isNull()) {
+        if ("folderPath".equals(key)) {
```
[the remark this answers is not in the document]
rrama: Already resolved: the `"folderPath"` literal was extracted to `LsFolderSettingsKeys.FOLDER_PATH` and the `if` checks now compare against the constant, so no string literal remains. Resolving.

tests/src/test/java/io/snyk/eclipse/plugin/preferences/HTMLSettingsPreferencePageTest.java
rrama: Done — replaced with a short present-tense note.

tests/src/test/java/io/snyk/eclipse/plugin/preferences/HTMLSettingsPreferencePageTest.java
rrama: Removed the line references.

tests/src/test/java/io/snyk/eclipse/plugin/preferences/HTMLSettingsPreferencePageTest.java
rrama: Removed.

rrama (commented):
LGTM, but I cannot do the manual testing due to IDE-2168.

basti-snyk (approved):

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/HTMLSettingsPreferencePage.java
```diff
@@ -78,6 +79,18 @@ public Object function(Object[] arguments) {
           if (parseAndSaveConfig(jsonString)) {
             SnykLanguageServer.promptToRestartEclipseForNewCli();
           }
+          // The dialog auto-saves through this bridge (e.g. "Reset overrides" is a commit point
+          // that fires immediately, not on Apply/OK). Persisting prefs is not enough — the LS only
+          // sees the change once didChangeConfiguration is sent. Push it now so resets and
+          // auto-saves take effect without the user pressing Apply (matches IntelliJ/VS Code).
+          SnykExtendedLanguageClient lc = SnykExtendedLanguageClient.getInstance();
+          if (lc != null) {
+            CompletableFuture.runAsync(lc::updateConfiguration)
```
basti-snyk, line 88: **[AI review]** Should fix: this `updateConfiguration()` push sits *after* the `if (parseAndSaveConfig(jsonString))` block closes, so it runs even when `parseAndSaveConfig` returned `false` (malformed JSON → nothing persisted). That sends a `didChangeConfiguration` carrying the previous, stale state on a failed save. Moving the push inside the success branch fires it only when the save actually applied.

plugin/src/main/java/io/snyk/languageserver/LsFolderSettingsKeys.java
```diff
@@ -9,10 +9,12 @@ private LsFolderSettingsKeys() {
 	public static final String BASE_BRANCH = "base_branch";
 	public static final String PREFERRED_ORG = "preferred_org";
 	public static final String ADDITIONAL_PARAMETERS = "additional_parameters";
-	public static final String ADDITIONAL_ENV = "additional_env";
+	/** LS/JS wire key for the folder-scoped additional environment override. */
+	public static final String ADDITIONAL_ENVIRONMENT = "additional_environment";
```
basti-snyk, line 13: **[AI review]** This changes the *wire* key, not just the Java identifier — the value goes from `"additional_env"` to `"additional_environment"`. Looks like a correctness fix (it now matches the global `LsKey` value and the shared FOLDER_RESET_FIELDS), but worth confirming: (1) it lands in lockstep with the snyk-ls side, and (2) whether any folder config persisted under the old `"additional_env"` key needs a one-time migration.

basti-snyk (commented):
**AI verification review (basti-snyk)** — 4-lens pass (semantic / adversarial / security / code-review) against the PR diff. CI is green, and the risky parts check out: null-first ordering in `processFolderConfig`, the `FOLDER_PATH` skip, the SCANNING_MODE boolean accept, and folder-vs-global scope isolation are all correct and well tested.

Two items worth addressing (inline comments):
- **Should fix** — LS push fires even when the save failed (bridge).
- **Confirm** — `additional_env` → `additional_environment` is a wire-key change.

Non-blocking suggestions:
- **Double `updateConfiguration` on Apply/OK:** `performOk` already pushes config, and the new bridge push also fires via `getAndSaveIdeConfig()`, so OK triggers two concurrent `didChangeConfiguration`. Idempotent but redundant — consider a single shared push path. (`performOk`’s own push also lacks an `exceptionally` handler.)
- **Silent null-element drop:** the array branch uses `.filter(el -> !el.isNull())` with no log, so a configured parameter can vanish without a trace — consider logging the skip like the sibling branches do.
- **Stale client on restart:** `lc` is captured before the async runs; re-resolving `getInstance()` inside the lambda avoids pushing to a disposed client if the LS restarts mid-flight.
- **Dead assertion:** the `ACTIVATE_SNYK_CODE_SECURITY` not-promoted check can’t fail (folder-only key isn’t in the global registry), so it adds no real coverage.

_Comment-only; no approval implied._

PR #429, opened 2026-06-19
feat: clear global overrides on Project Defaults reset [IDE-2149]

nick-y-snyk:
## What

Implements the Eclipse side of **IDE-2149** — global ("Project Defaults") reset overrides in the HTML settings page (sibling of the per-folder reset, IDE-1945).

Honors inbound global resets from the LS: when `$/snyk.configuration` sends a global setting as `{value:null, changed:true}`, Eclipse drops the persisted preference value and clears explicit-change tracking, so the override is not re-asserted on the next outbound sync/reconnect.

## Changes
- `SnykExtendedLanguageClient.persistGlobalSettings` — on a `null` + `changed:true` reset, `removePref(prefKey)` + `clearExplicitlyChanged(prefKey)` instead of skipping; non-null persist behavior unchanged. (Inbound still never *marks* settings as explicitly changed.)
- `Preferences.removePref` — remove a persisted value (falls back to default).
- `HTMLSettingsPreferencePage` — form-save null path also removes the persisted value so the next outbound sends `changed:false`.
- `InMemoryPreferenceStore` (test fake) — implement `remove`.
- Tests — inbound null reset clears value + explicit-changed; `null`+`changed:false` ignored; **re-push guard** in `LsConfigurationUpdaterTest` (`changed:false` after reset).

## Notes
- Mockito-based tests (`SnykExtendedLanguageClientTest`, `LsConfigurationUpdaterTest`) could not run locally — the inline mock-maker fails under JDK 21 inside the Tycho/Equinox surefire runtime. They will run in **CI**. PMD and compilation are clean locally; non-mock tests (`HTMLSettingsPreferencePageTest` 25/25, `PreferencesTest`, `FolderConfigSettingsTest`) pass.
- Requires the **snyk-ls** IDE-2149 PR for end-to-end behavior; safe to merge independently.


plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -397,6 +397,12 @@ private void persistGlobalSettings(Map<String, ConfigSetting> settings) {
 				var setting = entry.getValue();
 				if (setting.getValue() != null) {
 					prefs.store(registryEntry.prefKey, registryEntry.inboundDeserializer.apply(setting.getValue()));
+				} else if (Boolean.TRUE.equals(setting.getChanged())) {
+					// Global reset: LS Unset the user:global override and pushed back
+					// {value:null, changed:true}. Drop the persisted override and clear
+					// explicit-changed tracking so we don't re-assert it on the next sync.
+					prefs.removePref(registryEntry.prefKey);
```
basti-snyk: **Should Fix — two reset paths can be silently undone.** This is the core reset logic; two interactions can defeat it:

1. **Severity-filter partial reset re-asserts siblings.** The four severity-filter keys share an OR'd outbound `changed` flag in `LsConfigurationUpdater`/`LsSettingsRegistry`. If the LS pushes `{value:null, changed:true}` for only a subset of the four, this loop clears just those keys, but the un-reset siblings keep their persisted value and explicit-changed entry — and because the outbound `changed` is OR'd across the co-key group, all four are still sent `changed=true` on the next sync, re-asserting the overrides the reset was meant to clear. Worth confirming what subset the LS can emit, and handling the co-key group atomically if it can send a strict subset.

2. **Env-var reseed resurrects reset endpoint/org.** `ENDPOINT`/`ORGANIZATION` are non-encrypted registry entries reachable by this reset path, but the `Preferences` constructor unconditionally re-seeds them from `SNYK_API`/`SNYK_ORG` on every restart. So a reset of those keys is silently undone on the next plugin start when those env vars are set.

— AI review

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -397,6 +397,12 @@ private void persistGlobalSettings(Map<String, ConfigSetting> settings) {
 				var setting = entry.getValue();
 				if (setting.getValue() != null) {
 					prefs.store(registryEntry.prefKey, registryEntry.inboundDeserializer.apply(setting.getValue()));
+				} else if (Boolean.TRUE.equals(setting.getChanged())) {
```
basti-snyk: **Suggestion — code and comment disagree about the trigger.** The branch fires purely on `value==null && changed==true`, but the comment (and `removePref`'s javadoc) describe this as an "org-scope global" reset. `ConfigSetting.originScope` is deserialized from the wire yet never read here, so any `{value:null, changed:true}` — regardless of scope — performs the destructive override removal. Either gate on `"org".equals(setting.getOriginScope())`, or drop the "org-scope" wording so the code and its docs agree about what triggers the removal. — AI review

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
```diff
@@ -451,6 +451,12 @@ private void persistGlobalSettings(Map<String, ConfigSetting> settings) {
...
+					// Global reset: LS Unset the user:global override and pushed back
+					// {value:null, changed:true}. Drop the persisted override and clear
+					// explicit-changed tracking so we don't re-assert it on the next sync.
+					prefs.removePref(registryEntry.prefKey);
```
basti-snyk: **Should Fix — a global reset can silently wipe an env-seeded org/endpoint for the session.**

This re-review covers commits added since the 2026-06-23 review. New observation on this reset branch:

`ORGANIZATION_KEY` and `ENDPOINT_KEY` are seeded from the `SNYK_CFG_ORG` / `SNYK_API` environment variables in the constructor (`Preferences.java:160-168`) via `store(...)`, so they land in `insecurePreferences` as plain values that are indistinguishable from user:global overrides. `ORGANIZATION` is a non-encrypted, org-scope global setting in `LsSettingsRegistry`, so when the LS pushes `{value:null, changed:true}` for it, this branch calls `removePref(ORGANIZATION_KEY)`. After removal, `getPref` falls back to the store default (`""`) rather than the env value, and nothing re-seeds from the environment until Eclipse restarts.

Consequence: a user who configures their org/endpoint via env var and then receives a global reset would have the IDE send an empty org/endpoint outbound and scan against the wrong (default) org for the rest of the session.

Consider making env-seeded values reset-immune, or re-applying the env seed after `removePref` when an env var is present for that key.

— AI review

basti-snyk (approved):

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
```diff
@@ -29,7 +30,19 @@ LspConfigurationParam buildConfigurationParam() {
 		Preferences preferences = Preferences.getInstance();
 		Map<String, ConfigSetting> settings = new LinkedHashMap<>();
 
+		// Drain the one-shot reset queue once. For each queued key we emit
+		// {value:null, changed:true} — the only signal that makes snyk-ls Unset
+		// its user:global override — overriding the normal per-key emit below.
+		Set<String> pendingResets = preferences.consumePendingResets();
```
basti-snyk: **Should Fix — the one-shot reset can be silently lost because it is drained before delivery is confirmed.**

`consumePendingResets()` clears the queue here, inside `buildConfigurationParam()` — a method with more than one caller. The form-save push that is meant to carry the reset is fired fire-and-forget (`CompletableFuture.runAsync(lc::updateConfiguration).exceptionally(log)`). Two ways the drained `{value:null, changed:true}` never reaches snyk-ls:

1. If `configurationChanged()` reaches this build but the following `didChangeConfiguration` throws / the server is connected-but-unhealthy, the queue is already empty and nothing retries — the reset is dropped.
2. `getInitializationOptions()` at LS (re)start also calls `buildConfigurationParam()`, so a startup racing the form-save push can drain the reset into an init-options tree; that path swallows exceptions and returns `null`, discarding the reset.

The class comment argues transience is safe because restart re-emits the default with `changed=false` — but per this PR's own premise, only `{null, changed:true}` makes snyk-ls Unset a `user:global` override, so a dropped push is not self-healing. Consider draining only after a successful send (pass the drained set into the single outbound path), or re-queue on failure.

— AI review

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
```diff
@@ -29,7 +30,19 @@ LspConfigurationParam buildConfigurationParam() {
...
+		if (pendingResets == null) {
```
basti-snyk: **Suggestion — dead defensive guard.** `consumePendingResets()` always returns a non-null `new HashSet<>(...)`, so this `null` check and the `Collections.emptySet()` fallback are unreachable. Global project rules discourage error handling for impossible scenarios — this can be deleted (and `pendingResets` used directly).

— AI review

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
```diff
@@ -339,6 +343,21 @@ public final void store(String key, String value) {
 		}
 	}
 
+	/**
+	 * Removes the stored value for {@code key} so that subsequent reads fall back
+	 * to the preference store default. Used by the inbound global-reset path: when
```
basti-snyk: **Suggestion — Javadoc describes the wrong (inbound) caller.** This says `removePref` is "Used by the inbound global-reset path: when the language server pushes back {value:null, changed:true}...". But inbound null is deliberately *unsupported* — `SnykExtendedLanguageClient` skips it and `snykConfigurationInboundNullValueIsIgnored` locks that in. The only real caller is the **outbound** form-save path in `HTMLSettingsPreferencePage`. The doc inverts the shipped design and will mislead future readers; please point it at the outbound form-save/pending-reset path.

— AI review

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
nick-y-snyk: Fixed: moved `consumePendingResets()` into `configurationChanged()` so it drains only after confirming a live LS connection, preventing `getInitializationOptions()` from stealing the reset queue. `buildConfigurationParam()` now takes the set as a parameter; the no-arg overload passes `emptySet()` for the init path.

plugin/src/main/java/io/snyk/languageserver/LsConfigurationUpdater.java
nick-y-snyk: Already addressed — the null guard was removed as part of fixing #3511053370: `consumePendingResets()` no longer lives in `buildConfigurationParam()`; the set is now passed in by the caller, which always provides a non-null set.

plugin/src/main/java/io/snyk/eclipse/plugin/preferences/Preferences.java
nick-y-snyk: Fixed: updated the Javadoc to describe the actual outbound form-save/pending-reset path rather than the inbound path.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Two parts: (1) Severity-filter co-key group — confirmed the LS always emits all four severity keys together in a global reset, so partial-reset re-assertion can't happen. (2) Env-var reseed — fixed: `removePref()` now re-seeds `ENDPOINT_KEY`/`ORGANIZATION_KEY` from `SNYK_API`/`SNYK_CFG_ORG` env vars if present, so a reset doesn't leave the session with empty org/endpoint when env vars are set.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Dropped the 'org-scope' wording from comments — the trigger is purely `value==null && changed==true` regardless of `originScope`, and the code/comments now agree on that.

plugin/src/main/java/io/snyk/languageserver/protocolextension/SnykExtendedLanguageClient.java
nick-y-snyk: Fixed: `removePref()` now re-seeds `ENDPOINT_KEY`/`ORGANIZATION_KEY` from `SNYK_API`/`SNYK_CFG_ORG` env vars after removal, so the session retains the env-var value rather than falling back to empty string.

PR #438, opened 2026-07-09
fix: fix html styling to match native theme and stop flickering during scans.

andrewrobinsonhodges-snyk:
### Description

Before:
Screenshot 2026-07-08 at 17 17 07

After:
Screenshot 2026-07-09 at 09 42 15

### Checklist

- [ ] Read and understood the [Code of Conduct](https://github.com/snyk/snyk-eclipse-plugin/blob/main/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/snyk/snyk-eclipse-plugin/blob/main/CONTRIBUTING.md).
- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

### DIFF
--- a/plugin/src/main/java/io/snyk/eclipse/plugin/html/BaseHtmlProvider.java
+++ b/plugin/src/main/java/io/snyk/eclipse/plugin/html/BaseHtmlProvider.java
@@ -5,6 +5,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Random;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -40,7 +41,7 @@ public class BaseHtmlProvider {
 	// user switches Eclipse light↔dark at runtime those caches go stale. A global generation counter is
 	// bumped on theme change; each instance lazily clears its caches when it sees a newer generation, so
 	// the next render re-resolves foreground/border/etc. from the new theme (not just the background).
-	private static volatile int themeGeneration;
+	private static final AtomicInteger themeGeneration = new AtomicInteger();
 	private volatile int cacheGeneration;
 
 	public static void setIdeBackgroundColorHex(String hex) {
@@ -53,7 +54,7 @@ public class BaseHtmlProvider {
 
 	/** Signals that the Eclipse theme changed; cached theme colors are re-resolved on the next render. */
 	public static void invalidateThemeCaches() {
-		themeGeneration++;
+		themeGeneration.incrementAndGet();
 	}
 
 
@@ -409,7 +410,7 @@ public class BaseHtmlProvider {
 		if (Preferences.getInstance().isTest()) {
 			return;
 		}
-		int currentGeneration = themeGeneration;
+		int currentGeneration = themeGeneration.get();
 		if (cacheGeneration == currentGeneration) {
 			return;
 		}
--- a/plugin/src/main/java/io/snyk/eclipse/plugin/html/BaseHtmlProvider.java
+++ b/plugin/src/main/java/io/snyk/eclipse/plugin/html/BaseHtmlProvider.java
@@ -28,13 +28,26 @@ public class BaseHtmlProvider {
 	private final Map<String, String> colorCache = new HashMap<>();
 	private String nonce = "";
 
+	// The actual themed view background, captured from an SWT control that the e4 CSS engine has
+	// styled (see SnykToolView). This is the true "Eclipse grey" for the view content area — the tab
+	// color keys used below resolve to the lighter tab-strip grey, which does not match adjacent views.
+	// Shared statically so every provider instance (tree/summary/detail) resolves the same background.
+	private static volatile String ideBackgroundColorHex;
+
+	public static void setIdeBackgroundColorHex(String hex) {
+		ideBackgroundColorHex = hex;
+	}
+
+	public static String getIdeBackgroundColorHex() {
+		return ideBackgroundColorHex;
+	}
+
 
 	// Eclipse theme color keys
 	private static final String THEME_INACTIVE_TAB_BG = "org.eclipse.ui.workbench.INACTIVE_TAB_BG_START";
 	private static final String THEME_ACTIVE_TAB_KEYLINE = "org.eclipse.ui.workbench.ACTIVE_TAB_OUTER_KEYLINE_COLOR";
 	private static final String THEME_ACTIVE_TAB_SELECTED_TEXT = "org.eclipse.ui.workbench.ACTIVE_TAB_SELECTED_TEXT_COLOR";
 	private static final String THEME_ACTIVE_TAB_TEXT = "org.eclipse.ui.workbench.ACTIVE_TAB_TEXT_COLOR";
-	private static final String THEME_ACTIVE_NOFOCUS_TAB_BG = "org.eclipse.ui.workbench.ACTIVE_NOFOCUS_TAB_BG_START";
 	private static final String THEME_ACTIVE_TAB_BG_END = "org.eclipse.ui.workbench.ACTIVE_TAB_BG_END";
 	private static final String THEME_ACTIVE_HYPERLINK = "ACTIVE_HYPERLINK_COLOR";
 	private static final String THEME_DARK_BACKGROUND = "org.eclipse.ui.workbench.DARK_BACKGROUND";
@@ -182,12 +195,16 @@ public class BaseHtmlProvider {
 
 		boolean dark = Boolean.TRUE.equals(isDarkTheme());
 
+		// The true Eclipse view background. Both --ide-background-color and --background-color resolve to
+		// this single value so every panel (tree/summary/detail) matches each other and the adjacent views.
+		String ideBg = getIdeBackgroundHex();
+
 		// Replace CSS variables with actual color values
 		htmlStyled = htmlStyled.replace(CSS_VAR_TEXT_COLOR, getColorAsHex(THEME_ACTIVE_TAB_SELECTED_TEXT, "#000000"));
 		htmlStyled = htmlStyled.replace(CSS_VAR_DIMMED_TEXT_COLOR, getColorAsHex(THEME_ACTIVE_TAB_TEXT, "#4F5456"));
 
-		htmlStyled = htmlStyled.replace(CSS_VAR_IDE_BACKGROUND_COLOR, getColorAsHex(THEME_ACTIVE_NOFOCUS_TAB_BG, DEFAULT_WHITE_COLOR));
-		htmlStyled = htmlStyled.replace(CSS_VAR_BACKGROUND_COLOR, getColorAsHex(THEME_ACTIVE_TAB_BG_END, DEFAULT_WHITE_COLOR));
+		htmlStyled = htmlStyled.replace(CSS_VAR_IDE_BACKGROUND_COLOR, ideBg);
+		htmlStyled = htmlStyled.replace(CSS_VAR_BACKGROUND_COLOR, ideBg);
 		htmlStyled = htmlStyled.replace(CSS_VAR_CODE_BACKGROUND_COLOR,
 				getColorAsHex(THEME_INACTIVE_TAB_BG, DEFAULT_SECTION_BG_COLOR));
 		htmlStyled = htmlStyled.replace(CSS_VAR_BUTTON_COLOR,
@@ -212,7 +229,7 @@ public class BaseHtmlProvider {
 		// Replace CSS variables used in LS-served HTML (settings page)
 		// Get Eclipse theme colors
 		String textColor = getColorAsHex(THEME_ACTIVE_TAB_SELECTED_TEXT, "#000000");
-		String bgColor = getColorAsHex(THEME_ACTIVE_TAB_BG_END, DEFAULT_WHITE_COLOR);
+		String bgColor = ideBg;
 		String inputBgColor = getColorAsHex(THEME_INACTIVE_TAB_BG, DEFAULT_SECTION_BG_COLOR);
 		String borderColor = getColorAsHex(THEME_ACTIVE_TAB_KEYLINE, DEFAULT_BORDER_COLOR);
 		String focusColor = getColorAsHex(THEME_ACTIVE_HYPERLINK, "#0066cc");
@@ -299,6 +316,11 @@ public class BaseHtmlProvider {
 		vsCodeVarMap.put("vscode-scrollbarSlider-background", inputBgColor);
 		vsCodeVarMap.put("vscode-scrollbarSlider-hoverBackground", adjustForHover(inputBgColor, dark));
 		vsCodeVarMap.put("vscode-scrollbarSlider-activeBackground", adjustBrightness(inputBgColor, dark ? 1.2f : 0.8f));
+		// Tree-view tooltips (.tree-tooltip) key off these. Without a mapping the CSS falls back to a
+		// hardcoded dark background (#252526), which is invisible/unreadable in the light theme.
+		vsCodeVarMap.put("vscode-editorHoverWidget-background", inputBgColor);
+		vsCodeVarMap.put("vscode-editorHoverWidget-foreground", textColor);
+		vsCodeVarMap.put("vscode-editorHoverWidget-border", borderColor);
 		htmlStyled = replaceRemainingCssVars(htmlStyled, vsCodeVarMap);
 
 		htmlStyled = htmlStyled.replace("${headerEnd}", "");
@@ -347,6 +369,20 @@ public class BaseHtmlProvider {
 
 		// CSS allows 3 decimal places of precision for calculations.
 		return String.format(Locale.ROOT, "%.3frem", targetSizePt / startingFontSizePt);
+	}
+
+	/**
+	 * Returns the true Eclipse view background as a hex string. Prefers the value captured from a
+	 * themed SWT control (set via {@link #setIdeBackgroundColorHex(String)}), which reflects the exact
+	 * grey the active theme paints view content with. Falls back to the tab-strip background key when no
+	 * control-derived value is available (e.g. before the view is created, or in test mode).
+	 */
+	public String getIdeBackgroundHex() {
+		String captured = ideBackgroundColorHex;
+		if (captured != null && !captured.isEmpty()) {
+			return captured;
+		}
+		return getColorAsHex(THEME_ACTIVE_TAB_BG_END, DEFAULT_WHITE_COLOR);
 	}
 
 	public String getColorAsHex(String colorKey, String defaultColor) {
--- a/plugin/src/main/java/io/snyk/eclipse/plugin/html/BaseHtmlProvider.java
+++ b/plugin/src/main/java/io/snyk/eclipse/plugin/html/BaseHtmlProvider.java
@@ -4,6 +4,7 @@ import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Random;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -25,12 +26,22 @@ import io.snyk.languageserver.protocolextension.messageObjects.PresentableError;
 
 public class BaseHtmlProvider {
 	private final Random random = new Random();
-	private final Map<String, String> colorCache = new HashMap<>();
+	// ConcurrentHashMap, not HashMap: the detail panel renders off the UI thread
+	// (BrowserHandler.updateBrowserContent → CompletableFuture.runAsync) against a shared singleton
+	// provider, so refreshThemeCachesIfStale's clear() can race getColorAsHex's computeIfAbsent.
+	private final Map<String, String> colorCache = new ConcurrentHashMap<>();
 	private String nonce = "";
 
 	// The native themed view background color from Eclipse.
 	// Shared statically so every provider instance (tree/summary/detail) resolves the same background.
 	private static volatile String ideBackgroundColorHex;
+
+	// Resolved theme colors are cached per instance (colorCache/colorRegistry/currentTheme). When the
+	// user switches Eclipse light↔dark at runtime those caches go stale. A global generation counter is
+	// bumped on theme change; each instance lazily clears its caches when it sees a newer generation, so
+	// the next render re-resolves foreground/border/etc. from the new theme (not just the background).
+	private static volatile int themeGeneration;
+	private volatile int cacheGeneration;
 
 	public static void setIdeBackgroundColorHex(String hex) {
 		ideBackgroundColorHex = hex;
@@ -38,6 +49,11 @@ public class BaseHtmlProvider {
 
 	public static String getIdeBackgroundColorHex() {
 		return ideBackgroundColorHex;
+	}
+
+	/** Signals that the Eclipse theme changed; cached theme colors are re-resolved on the next render. */
+	public static void invalidateThemeCaches() {
+		themeGeneration++;
 	}
 
 
@@ -178,6 +194,7 @@ public class BaseHtmlProvider {
 	}
 
 	public String replaceCssVariables(String html, boolean useRelativeFontSize) {
+		refreshThemeCachesIfStale();
 		// Build the CSS with the nonce
 		String nonce = getNonce();
 		String css = "<style nonce=\"" + nonce + "\">" + getCss() + "</style>";
@@ -381,6 +398,32 @@ public class BaseHtmlProvider {
 		return getColorAsHex(THEME_ACTIVE_TAB_BG_END, DEFAULT_WHITE_COLOR);
 	}
 
+	// Re-resolves this instance's cached theme colors if a theme change has been signalled since the
+	// last render, so subsequent getColorAsHex lookups reflect the current theme. Only re-fetches inside
+	// the branch (never taken until invalidateThemeCaches bumps the generation), so test renders — which
+	// have no running workbench — never call PlatformUI here.
+	private void refreshThemeCachesIfStale() {
+		// Test mode never resolves real colors (getColorAsHex short-circuits), and there is no workbench,
+		// so keep the PlatformUI call out of headless runs — self-enforcing rather than relying on the
+		// generation never being bumped in tests.
+		if (Preferences.getInstance().isTest()) {
+			return;
+		}
+		int currentGeneration = themeGeneration;
+		if (cacheGeneration == currentGeneration) {
+			return;
+		}
+		colorCache.clear();
+		ITheme freshTheme = PlatformUI.getWorkbench().getThemeManager().getCurrentTheme();
+		// Publish currentTheme/colorRegistry (both volatile) BEFORE the volatile cacheGeneration write, so
+		// a concurrent reader that observes the new generation is guaranteed to also see the new fields
+		// (happens-before via the cacheGeneration volatile). Writing the generation first would let a
+		// reader skip its own refresh and then read a stale registry.
+		currentTheme = freshTheme;
+		colorRegistry = freshTheme.getColorRegistry();
+		cacheGeneration = currentGeneration;
+	}
+
 	public String getColorAsHex(String colorKey, String defaultColor) {
 		if (Preferences.getInstance().isTest()) {
 			return "";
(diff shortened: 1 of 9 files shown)


plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
```diff
@@ -62,11 +104,171 @@ public void createPartControl(Composite parent) {
...
+	 * re-publish it and re-render each panel so they adopt the correct grey. Attempted twice to guard
+	 * against the CSS styling landing a cycle later than the first attempt; the work is idempotent, so
+	 * the second attempt is a no-op once the color has been applied.
+	 */
+	private void captureThemeBackgroundDeferred() {
+		if (rootComposite == null || rootComposite.isDisposed()) {
+			return;
+		}
+		rootComposite.getDisplay().asyncExec(this::applyThemeColorsIfChanged);
+		rootComposite.getDisplay().timerExec(300, this::applyThemeColorsIfChanged);
```
basti-snyk, line 180: **Should-Fix — theme background goes stale after a runtime theme switch.**

The IDE background is captured once at view creation via `asyncExec` + a fixed `timerExec(300, …)` best-effort re-read, and stored in the process-global `BaseHtmlProvider.ideBackgroundColorHex`. There is no subscription to the e4 CSS-styling / theme-change event.

Failure modes:
- If the CSS engine styles the view later than ~300 ms (slow startup / many views), the captured color is pre-theme and never corrected.
- If the user toggles Eclipse dark↔light while the view is open, `applyThemeColorsIfChanged` never re-runs, so panel/wrapper backgrounds and the sash divider keep the old color and mismatch adjacent views — regressing the live per-render resolution that existed before this PR and defeating the fix's own goal.

Root-cause fix: subscribe to the theme-styling/theme-change event and re-invoke `applyThemeColorsIfChanged` (which can then also re-render the panels via their `refreshTheme()`), keeping one deferred attempt only as a fallback rather than the primary mechanism.

— AI review

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/TreeViewBrowserHandler.java
```diff
@@ -7,15 +7,42 @@
...
-	private final BaseHtmlProvider htmlProvider = new BaseHtmlProvider();
+	// The feedback banner link (styles.css: .feedback-banner-link) uses --button-text-color, which the
+	// IDE resolves to the button foreground — black in the dark theme (readable on a light button, but
+	// not on the banner's dark gradient). Override it to the theme text color (white in dark, dark in
+	// light), matching the banner text. Injected via ${ideStyle}, which lands after the LS styles so it
+	// wins by cascade order.
+	private final BaseHtmlProvider htmlProvider = new BaseHtmlProvider() {
+		@Override
+		public String getCss() {
+			return ".feedback-banner-link { color: var(--text-color); }";
```
basti-snyk: **Should-Fix — this override silently no-ops if the tree HTML has no injection point.**

`getCss()` is only emitted by `replaceCssVariables`, which injects it where the HTML contains `${ideStyle}` (or the `data-ide-style` sentinel). If the language-server-served tree HTML does not contain that marker, the `.feedback-banner-link { color: var(--text-color); }` rule is never inserted and the readability fix does nothing — with no error.

Please confirm the live LS tree template actually contains the `${ideStyle}` placeholder, and add a test asserting the injected rule appears in the rendered tree output so this can't regress unnoticed.

— AI review

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SummaryBrowserHandler.java
```diff
@@ -34,15 +37,31 @@ public Object function(Object[] arguments) {
...
 	public void setDefaultBrowserText() {
-		browser.setText(StaticPageHtmlProvider.getInstance().getSummaryInitHtml());
+		lastSummary = null;
+		BrowserFlashGuard.setTextFlashSafe(browser, StaticPageHtmlProvider.getInstance().getSummaryInitHtml());
 	}
 
 	public void setBrowserText(String summary) {
-		browser.setText(StaticPageHtmlProvider.getInstance().getFormattedSummaryHtml(summary));
+		lastSummary = summary;
+		BrowserFlashGuard.setTextFlashSafe(browser,
```
basti-snyk: **Suggestion — summary panel hides on every render, unlike the tree.**

`updateSummary` runs `setTextFlashSafe` (hide → reveal on `completed`) on every call and has no unchanged-content guard. The tree handler was deliberately exempted from hiding (to avoid swallowing clicks) and given an `html.equals(lastRenderedHtml)` dedup guard because the LS re-pushes frequently with identical content.

`SnykExtendedLanguageClient.updateSummaryPanel` calls this repeatedly during a scan, often with identical summary text, so the summary panel can blink to the panel color on each push — reintroducing, inconsistently, the flicker this PR removes elsewhere. Consider mirroring the tree's approach: skip re-render when the summary is unchanged, and reconsider the hide-on-every-render for this panel.

— AI review

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
andrewrobinsonhodges-snyk: Addressed. Added an `IThemeManager` `CHANGE_CURRENT_THEME` listener that re-captures the themed background and re-renders each panel on a runtime light↔dark switch, plus a theme-generation counter in `BaseHtmlProvider` that invalidates the per-instance color caches so foreground/border colours re-resolve too (not just the background). The resize listener also re-invokes `applyThemeColorsIfChanged` as a fallback for the 'CSS styled later than ~300ms' case. — AI review

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SummaryBrowserHandler.java
```diff
@@ -34,15 +37,31 @@
 			}
 		};
 
+		BrowserFlashGuard.install(browser);
 		setDefaultBrowserText();
 	}
 
 	public void setDefaultBrowserText() {
-		browser.setText(StaticPageHtmlProvider.getInstance().getSummaryInitHtml());
+		lastSummary = null;
```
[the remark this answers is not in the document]
andrewrobinsonhodges-snyk: Fixed — `setDefaultBrowserText()` no longer assigns `null`; it uses a `hasSummary` boolean sentinel and `refreshTheme()` keys off that instead. — AI review

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SummaryBrowserHandler.java
andrewrobinsonhodges-snyk: Addressed. Added an unchanged-content guard to `SummaryBrowserHandler` (mirrors the tree's dedup) so identical summary pushes during a scan are skipped and no longer blink the panel. Extracted a private `renderSummary()` so `refreshTheme()` can force a re-render on theme change while `setBrowserText` keeps the dedup. Kept hide-on-genuine-change (flash-safe) — the summary isn't the click-critical panel, so a dark transition on a real update is preferable to a white flash. — AI review

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/TreeViewBrowserHandler.java
```diff
@@ -25,9 +52,30 @@
...
+	public void refreshTheme() {
+		if (browser == null || browser.isDisposed()) {
+			return;
+		}
+		if (lastRenderedHtml == null) {
+			setEmptyPage();
+			return;
+		}
+		String html = lastRenderedHtml;
+		lastRenderedHtml = null;
```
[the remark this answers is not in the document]
andrewrobinsonhodges-snyk: Fixed — `refreshTheme()` no longer nulls `lastRenderedHtml`. Extracted a shared private `render(html)` called by both `setBrowserText` (after its dedup check) and `refreshTheme` (bypassing it), so the null-assignment is gone. — AI review

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/TreeViewBrowserHandler.java
andrewrobinsonhodges-snyk: Confirmed + covered. The live LS tree template does contain the marker — `snyk-ls .../treeview/template/tree.html` has `${ideStyle}` immediately after the `{{.Styles}}` block. Added `TreeViewHtmlProviderTest` (3 tests) asserting the `.feedback-banner-link` override is injected when `${ideStyle}` is present and omitted when it isn't, so a `getCss()`/injection regression is caught. The runtime-served LS template itself can't be asserted from a plugin unit test, hence the manual confirmation. — AI review

plugin/src/main/java/io/snyk/eclipse/plugin/html/BaseHtmlProvider.java
```diff
@@ -349,6 +385,45 @@ private String getRelativeFontSize(int inputFontSizePt) {
...
+		// so keep the PlatformUI call out of headless runs — self-enforcing rather than relying on the
+		// generation never being bumped in tests.
+		if (Preferences.getInstance().isTest()) {
+			return;
+		}
+		int currentGeneration = themeGeneration;
+		if (cacheGeneration == currentGeneration) {
+			return;
+		}
+		colorCache.clear();
```
basti-snyk: **Should-Fix — `colorCache.clear()` runs before the registry swap, so a concurrent render can re-cache an old-theme color that then persists.**

The volatile-publish ordering below (fields before `cacheGeneration`) is correct for *new* readers, but `colorCache.clear()` happens on line 416, while `colorRegistry`/`currentTheme` aren't swapped until lines 422-423. A `getColorAsHex` → `computeIfAbsent` on another thread that runs in the window between the clear and the swap resolves against the **old** `colorRegistry` and repopulates `colorCache` with a pre-switch color. Once `cacheGeneration` is bumped (424), no further clear occurs for this generation, so that stale entry survives until the *next* theme change — the exact race the class comment on line 31 calls out.

Root-cause fix: swap `currentTheme`/`colorRegistry` first, then `colorCache.clear()` last (keeping the generation write last). That way any entry a concurrent compute inserted with the old registry is wiped, and post-clear computes all resolve against the new registry.

Narrow (needs two overlapping renders of the singleton provider — e.g. rapid detail-panel selections during a theme switch), but it defeats this PR's own goal, so worth tightening.

— AI review

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
```diff
@@ -62,11 +113,219 @@ public void createPartControl(Composite parent) {
...
+		};
+		themeManager.addPropertyChangeListener(themeChangeListener);
+	}
+
+	private void onThemeChanged() {
+		// Drop cached theme colors so panels re-resolve foreground/border/etc. (not just the
+		// background), then re-read the now-restyled background and re-render every panel. force=true
+		// because a theme change can alter foreground/accent colors while leaving the view background
+		// hex unchanged — the hex-equality gate below would otherwise skip the re-render.
+		BaseHtmlProvider.invalidateThemeCaches();
```
basti-snyk: **Should-Fix — theme toggle can paint new-theme foreground on the old background for up to ~300 ms.**

`onThemeChanged()` bumps the theme generation *immediately* here (`invalidateThemeCaches()`), which clears `colorCache` and makes the next `replaceCssVariables` re-resolve foreground/border/link colors from the **new** theme registry. But the new view background is only published later, inside the `timerExec(300, () -> applyThemeColors(true))` deferred pass (`applyThemeColors` → `setIdeBackgroundColorHex` at ~L212). `getIdeBackgroundHex()` therefore keeps returning the **old** background during that ~300 ms window.

Consequence: any render triggered in that window — e.g. an LS tree/summary push while a scan is running, or a node click — combines new-theme text with the old background (classic case: switching dark→light gives light-theme dark text on the still-dark background, unreadable until the deferred pass corrects it). The comment above intends "re-resolve foreground, not just background," but the background half is exactly what lags.

Root-cause fix: publish the background (read + `setIdeBackgroundColorHex`) at the same point you bump the generation — or don't bump the generation until the background is refreshed — so foreground and background always flip together. Keep the 300 ms forced re-render as the follow-up styling pass.

— AI review

plugin/src/main/java/io/snyk/eclipse/plugin/html/BaseHtmlProvider.java
andrewrobinsonhodges-snyk: Fixed — reordered `refreshThemeCachesIfStale`: swap `currentTheme`/`colorRegistry` first, then `colorCache.clear()`, then bump `cacheGeneration` last. A stale entry a concurrent compute inserted against the old registry before the swap is now wiped by the clear, and every post-clear compute resolves against the new registry. — AI review

plugin/src/main/java/io/snyk/eclipse/plugin/html/TreeViewHtmlProvider.java
rrama, line 1: Would it not be better to do a real fix in LS so other IDEs benefit and don't have to do the same fix? If so, please mark this as a temporary fix and raise a tech-debt Jira ticket to do the real fix after release.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
andrewrobinsonhodges-snyk: Fixed — moved `invalidateThemeCaches()` into the 300ms deferred pass so the foreground cache invalidation and the background publish (`setIdeBackgroundColorHex`) flip together. During the window renders stay fully on the old theme (consistent/readable) rather than new-foreground-on-old-background, and the delay also lets the e4 CSS engine restyle the view before we read the new background. — AI review

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BrowserFlashGuard.java
```diff
@@ -0,0 +1,63 @@
...
+	 * Hides the browser, loads the HTML, and schedules a safety reveal so the panel can never get stuck
+	 * hidden if the completed event does not fire.
+	 */
+	static void setTextFlashSafe(Browser browser, String html) {
+		if (browser == null || browser.isDisposed() || html == null) {
+			return;
+		}
+		browser.setVisible(false);
+		browser.setText(html);
+		browser.getDisplay().timerExec(SAFETY_REVEAL_MS, () -> reveal(browser));
```
rrama: AI written comment explaining my thoughts:

> The safety-reveal timers stack rather than coalesce. Each setTextFlashSafe call creates a fresh () -> reveal(browser) lambda, and SWT's timerExec cancels by Runnable instance, so a new call never cancels the previous timer. If update B hides + setTexts while update A's 1500ms timer is still pending, A's timer fires mid-load and reveals the browser before B's document is ready.
> 
> This matters more than it looks because completed after setText is known to be unreliable/platform-dependent (the reason this class has the safety timer at all — see the :46-47 javadoc). When completed doesn't fire, the 1500ms timer is the reveal path, so on the summary panel — which re-hides on every changed render during a scan — a rapid B-after-A sequence can reveal on A's stale timer while B is still loading. The tree isn't affected (hidden once at cold-open, never re-hidden).
> 
> Fix: store one reusable reveal Runnable per browser and timerExec(-1, revealRunnable) before rescheduling, so only the latest timer is ever live.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
```diff
@@ -159,17 +414,44 @@ private void drainPendingHtmlAsync() {
...
+		}
+		long now = System.currentTimeMillis();
+		if (firstPendingSince == 0) {
+			firstPendingSince = now;
+		}
+		if (now - firstPendingSince >= TREE_RENDER_MAX_WAIT_MS) {
+			display.timerExec(-1, treeRenderRunnable); // cancel any pending debounce
+			renderPendingTreeHtml();
+		} else {
+			display.timerExec(TREE_RENDER_DEBOUNCE_MS, treeRenderRunnable);
```
rrama, line 456: AI comment (same issue I spotted in the other file):

> **Debounce timer accumulates without canceling previous timers**
> 
> In the `else` branch at line 445, `display.timerExec(TREE_RENDER_DEBOUNCE_MS, treeRenderRunnable)` is called without first canceling the previously scheduled timer (via `timerExec(-1, treeRenderRunnable)`). SWT's `timerExec` does not implicitly replace a previous timer for the same Runnable — it adds a new one. So if `scheduleTreeRender` is called N times within the debounce window, N timers will fire. This is functionally harmless because `renderPendingTreeHtml` uses `pendingHtml.getAndSet(null)` and checks for null, making subsequent firings no-ops. However, adding a cancel before the schedule would be cleaner and avoid N-1 unnecessary timer callbacks.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BrowserHandler.java
```diff
@@ -176,30 +181,40 @@ public void updateBrowserContent(String issueId, String product) {
...
+		// re-runs replaceCssVariables and picks up a newly-resolved theme color. Rendering pre-resolved
+		// HTML here would freeze the error page at whatever color was current when it first displayed.
+		currentRenderer = () -> setErrorBrowserText(error);
+		BrowserFlashGuard.setTextFlashSafe(browser, new BaseHtmlProvider().getErrorHtml(error));
 	}
 
-	public void setBrowserText(String html) {
-		if (!browser.isDisposed()) {
-			browser.setText(html);
+	/** Re-renders the current content so it adopts a newly-resolved theme color. */
```
rrama, line 217: AI found nitpick:

> **refreshTheme on the detail panel re-fetches the issue description from the language server**
> 
> When `currentRenderer` is `() -> updateBrowserContent(issueId, product)` (set at `BrowserHandler.java:170`), calling `refreshTheme()` at line 218 re-invokes `updateBrowserContent`, which calls `SnykExtendedLanguageClient.getInstance().getIssueDescription(issueId)` on a background thread. This is a network/IPC call to the language server, not a cheap local re-render. The comment at line 206-208 explains this is intentional for the error case (to re-run `replaceCssVariables`), but the same pattern applies to the issue detail case. On a theme change, this means every displayed issue detail triggers a full round-trip to the LS. This is acceptable for a rare user action (theme toggle) but worth being aware of if theme-change frequency increases.

Happy to just have it documented in a comment somewhere.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BrowserHandler.java
```diff
@@ -176,30 +181,40 @@ public void updateBrowserContent(String issueId, String product) {
 			}
 			final var browserContent = htmlProvider.replaceCssVariables(htmlContent);
 			Display.getDefault().syncExec(() -> {
```
rrama, line 187: AI spotted a pre-exisiting nitpick:

> **syncExec in updateBrowserContent blocks the ForkJoinPool common thread**
> 
> At `BrowserHandler.java:183`, `Display.getDefault().syncExec(...)` is called inside a `CompletableFuture.runAsync` lambda, which by default runs on the ForkJoinPool common pool. `syncExec` blocks the calling thread until the UI thread processes the runnable. If the UI thread is busy (e.g., processing a theme change or a long paint), this blocks a common pool thread. With enough concurrent `updateBrowserContent` calls, this could exhaust the common pool. This is a pre-existing pattern (not introduced by this PR), and in practice the UI thread processes syncExec quickly, but it's worth noting as a potential issue under heavy load.

Probably not worth changing in this PR though.

rrama (approved):
A few bits could be improved to further reduce flickering, but nothing blocking if you wanted to do any of it as a follow-up.

There does appear to be a lack of testing though, so I had AI write me a list of what it thinks would be good to add (unread by me):

> **Missing tests, ranked by impact**
> 1. Tree dedup guard + refreshTheme bypass
> Gap: The html.equals(lastRenderedHtml) skip in TreeViewBrowserHandler.java:48-51 and the deliberate bypass in refreshTheme (TreeViewBrowserHandler.java:59-68) have no test — even though the near-identical summary guard got SummaryBrowserHandlerTest.java. Difficulty: Medium. Not testable as-is: the browser == null guard runs before the dedup (TreeViewBrowserHandler.java:42-51), and render calls browser.setText directly (TreeViewBrowserHandler.java:70-73) with no overridable seam. Needs a small refactor first. Notes: Copy the exact recipe already used for the summary handler — make render package-private/overridable (see the comment at SummaryBrowserHandler.java:75-77), move the null-browser guard so lastRenderedHtml bookkeeping still runs, then subclass with super(null) and count render calls. Assert: first call renders; identical repeat skips; changed HTML renders; refreshTheme re-renders identical HTML; refreshTheme with lastRenderedHtml == null calls setEmptyPage. Use Preferences.getTestInstance(...).setTest(true) in `@BeforeEach`.
> 
> 2. TreeViewBrowserHandler.escapeJsSingleQuotedString
> Gap: JS-injection escaping into browser.evaluate at TreeViewBrowserHandler.java:82-89 — a safety-relevant routine, no test. Difficulty: Easy. Already static and package-private; no SWT, no workbench needed. Notes: Assert each escape independently (backslash, single-quote, \n, \r, U+2028, U+2029) plus a combined string. One-file test, highest value-per-effort.
> 
> 3. createDividerRgb — luminance branch + clamping
> Gap: Pure divider-color function in SnykToolView.java (lightens dark by +32, darkens light by −28, clamps 0–255). No coverage. Difficulty: Medium. Pure, but private on the ViewPart; make it package-private static (uses no instance state) or move to a util. Notes: Assert dark input (luminance < 0.5) → +32/channel; light input → −28; near-black clamps at 0; near-white clamps at 255; a value straddling the 0.5 threshold picks the correct branch.
> 
> 4. toHex
> Gap: Pure hex formatting in SnykToolView.java. No coverage. Difficulty: Medium. Same visibility caveat as no. 3. Notes: Extract/make package-private alongside createDividerRgb. Assert zero-padding (new RGB(1,2,3) → #010203), null → null, full white/black.
> 
> 5. Debounce + max-wait ceiling
> Gap: scheduleTreeRender/renderPendingTreeHtml in SnykToolView.java — the starvation-prevention "render immediately once pending > TREE_RENDER_MAX_WAIT_MS" branch. Also where Bug Catcher flagged the else-branch not cancelling its timer. Difficulty: Hard as written. Coupled to Display.timerExec and System.currentTimeMillis() on a full ViewPart, with no clock/scheduler seam. Notes: Don't test in place. Extract the decision into a pure helper — boolean shouldRenderNow(long now, long firstPendingSince) or a tiny TreeRenderDebouncer with an injected LongSupplier clock — then unit-test the MAX_WAIT_MS boundary. This also makes the flagged cancel-before-reschedule bug inspectable/fixable.
> 
> 6. Theme-cache invalidation / generation counter
> Gap: invalidateThemeCaches/refreshThemeCachesIfStale + generation counter in BaseHtmlProvider.java — core of the runtime theme-switch fix and subject of the clear/swap-ordering race. Difficulty: Split. The test-mode short-circuit is Easy to cover (set setTest(true), bump generation, assert no throw / cache untouched). The interesting registry-swap ordering is Hard: refreshThemeCachesIfStale calls PlatformUI.getWorkbench().getThemeManager() and there's no live workbench (BaseTreeNodeTest.java:160-170). Notes: You could mockStatic(PlatformUI.class) — the pattern exists (ResourceUtilsTest.java:51) — to stub getWorkbench().getThemeManager().getCurrentTheme() and assert the swap-then-clear order and generation bump. Cover the test-mode branch for sure; treat the mocked-workbench swap-ordering test as optional (it verifies the ordering contract, but against a stub, not a real registry).
> 
> 7. BrowserFlashGuard
> Gap: No tests for BrowserFlashGuard.java. Difficulty: Easy for the null/disposed no-op guards; Hard for the meaningful hide→setText→reveal-on-completed + 1.5s safety-reveal behavior (needs a real Browser + event loop, unavailable in this harness). Notes: Low value-per-effort. At most assert install(null), setTextFlashSafe(null, …), scheduleSafetyReveal(null) don't throw. The timer/reveal logic is genuinely integration-test territory; don't fake it.
> 
> Recommended set to actually implement now: 1, 2, 3, 4, plus the test-mode branch of 6 — all achievable in this fragment-JUnit harness (some with a tiny visibility/seam refactor). 5 should be gated on extracting the debounce logic. 7's real behavior and #6's swap-ordering are integration-test territory given there's no live workbench here.

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BrowserFlashGuard.java
andrewrobinsonhodges-snyk: Good catch — fixed. `BrowserFlashGuard` now uses a single reusable reveal `Runnable` per browser (stored in widget data) and `timerExec(-1, revealTask)` cancels any pending safety-reveal before rescheduling, so only the latest timer is ever live. A stale timer from update A can no longer fire mid-load of update B and reveal it early — the real risk on the summary panel that re-hides on every changed render (the tree is unaffected, hidden only at cold-open). — AI review

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/SnykToolView.java
andrewrobinsonhodges-snyk: Fixed — `scheduleTreeRender` now calls `timerExec(-1, treeRenderRunnable)` before scheduling, so rapid calls coalesce into a single debounce timer instead of stacking N callbacks. (Agreed the extras were harmless no-ops via `getAndSet(null)`; this just avoids them.) — AI review

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BrowserHandler.java
andrewrobinsonhodges-snyk: Documented — added a comment on `updateBrowserContent`'s `currentRenderer` noting that `refreshTheme` re-runs it (re-fetching the description from the LS), that this is acceptable only because a theme switch is rare, and the future optimization (cache the raw HTML and re-run only `replaceCssVariables`) if theme-change frequency increases. — AI review

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BrowserHandler.java
andrewrobinsonhodges-snyk: Agreed — pre-existing and out of scope for this PR. Folding it into the same off-thread tech-debt ticket as the `refreshThemeCachesIfStale` theme-access item (#4956248032). Left the thread open for you to close. — AI review

plugin/src/main/java/io/snyk/eclipse/plugin/html/TreeViewHtmlProvider.java
andrewrobinsonhodges-snyk: Agreed the real fix belongs in the LS. Marked `TreeViewHtmlProvider` as a temporary plugin-side workaround (class javadoc + a `TODO` on `getCss`), noting the banner reuses the shared `--button-text-color` and that the LS should give the banner link its own readable color variable so every IDE benefits and none needs this override. Leaving this thread open pending a tech-debt Jira ticket for the LS change. — AI review

andrewrobinsonhodges-snyk: Raised as IDE-2234 (https://snyksec.atlassian.net/browse/IDE-2234) for the snyk-ls banner-link color variable; the plugin-side override is now marked `TODO(IDE-2234)` to be removed once the LS fix lands. — AI review

plugin/src/main/java/io/snyk/eclipse/plugin/views/snyktoolview/BrowserHandler.java
andrewrobinsonhodges-snyk: Tracked as IDE-2235 (https://snyksec.atlassian.net/browse/IDE-2235) alongside the off-thread theme-access refactor. — AI review

PR #442, opened 2026-07-23
chore: sync settings-fallback.html from snyk-ls

team-ide-user:
Automatic sync of `settings-fallback.html` triggered by [snyk/snyk-ls@920eb24](https://github.com/snyk/snyk-ls/commit/920eb24fdb6677d8767aa86ca6153242ca80f535).

This PR was opened by the [distribute-fallback-html](https://github.com/snyk/snyk-ls/blob/main/.github/workflows/distribute-fallback-html.yaml) workflow in snyk-ls.

## What changed

`settings-fallback.html` is the settings page displayed before the Language Server binary is available. It is maintained in [snyk/snyk-ls](https://github.com/snyk/snyk-ls) and must be kept in sync with the copy in each IDE plugin. This repo's `resource-check` CI workflow verifies that the local copy matches the snyk-ls main branch; this PR restores that sync.

## Merge instructions

Review and merge when ready. No manual testing is required beyond confirming that `resource-check` passes on this PR.

### DIFF
--- a/plugin/src/main/resources/ui/html/settings-fallback.html
+++ b/plugin/src/main/resources/ui/html/settings-fallback.html
@@ -59,6 +59,14 @@
       --border-color: var(--vscode-panel-border, #454545);
       --focus-border: var(--vscode-focusBorder, #007acc);
 
+      /* Buttons */
+      --button-background: var(--vscode-button-background, #0e639c);
+      --button-foreground: var(--vscode-button-foreground, #ffffff);
+      --button-hover-background: var(--vscode-button-hoverBackground, #1177bb);
+      --button-secondary-background: var(--vscode-button-secondaryBackground, #3a3d41);
+      --button-secondary-foreground: var(--vscode-button-secondaryForeground, #ffffff);
+      --button-secondary-hover-background: var(--vscode-button-secondaryHoverBackground, #45494e);
+
       /* Other */
       --border-radius: 4px;
       --border-radius-small: 2px;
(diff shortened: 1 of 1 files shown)


rrama (approved):

rrama:
Checked the review bot's two findings ([1](https://github.com/snyk/snyk-eclipse-plugin/pull/442#issuecomment-5058527809), [2](https://github.com/snyk/snyk-eclipse-plugin/pull/442#issuecomment-5060501096)) against the actual bridge/theming code — both false positives.

**"Logic Error" [major]** — claims passing the JS `done` function to `window.__ideExecuteCommand__` breaks because SWT `BrowserFunction` can't marshal function objects across the JS↔Java boundary, so the callback is received as null. But `ExecuteCommandBridge.java`'s `buildClientScript()` never passes the function itself across the bridge — it stores `done` in a client-side `window.__ideCallbacks__` map keyed by a generated `callbackId` string, and only strings (`command`, `argsJson`, `callbackId`) are passed to the native `__ideExecuteCommandBridge__` BrowserFunction. The Java side calls back into JS via `browser.evaluate(...)`, invoking the callback through that same client-side map. This callbackId-indirection is the whole point of the design (same pattern used in the VS Code, IntelliJ, and Visual Studio bridges) — specifically so a function object never has to cross the native boundary.

**"Theme Variable Mismatch"/"Theme Integration Failure" [minor, posted twice]** — claims the new `--button-background`/`--button-foreground` variables won't match `BaseHtmlProvider.java`'s `CSS_VAR_LS_BUTTON_BACKGROUND_COLOR`/`CSS_VAR_LS_BUTTON_FOREGROUND` constants, so the button falls back to hardcoded colors. But those *primary* button variables are declared in `:root` and never actually consumed anywhere in `settings-fallback.html` — the visible "Clear credentials" button uses the *secondary* variant (`background-color: var(--button-secondary-background)`, etc.), and `BaseHtmlProvider.java` has exact matching constants for those (`CSS_VAR_LS_BUTTON_SECONDARY_BACKGROUND`, `CSS_VAR_LS_BUTTON_SECONDARY_FOREGROUND`, `CSS_VAR_LS_BUTTON_SECONDARY_HOVER_BACKGROUND`). Theming works correctly for the button that actually renders.

No change needed on either.

